PR #3, opened 2022-08-16
Avoid sending tokens to token contract, add tests for sending tokens …

grandizzy:
…to token contract and 0x address


MikeHathaway (approved):
LGTM

PR #4, opened 2022-08-17
Remove token constructor, add some more tests

grandizzy:


src/Token.sol
```diff
@@ -13,10 +13,6 @@ import { ERC1967Proxy } from "@oz/proxy/ERC1967/ERC1967Proxy.sol";
 
 
 contract AjnaToken is Initializable, ERC20Upgradeable, ERC20BurnableUpgradeable, ERC20PermitUpgradeable, ERC20VotesUpgradeable, OwnableUpgradeable, UUPSUpgradeable {
-    /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor() {
```
MikeHathaway, line 17: What is the purpose of removing the constructor? This setup is explicitly recommended by OZ

grandizzy: Hm, according to https://www.youtube.com/watch?v=Vt20jCu8OC8 it should be removed

test/Token.t.sol
```diff
@@ -141,8 +176,9 @@ contract TokenTest is Test {
         tokenProxyV2.setTestVar(100);
         assertEq(tokenProxyV2.testVar(), 100);
 
-        // TODO: check previous state no longer accessible...
-        // assertEq(tokenProxyV1.totalSupply(), 1_000_000_000 * 10 ** tokenProxyV1.decimals());
+        // FIXME: shouldn't these be 1bil as it used to be in v1?
+        assertEq(tokenProxyV2.totalSupply(),            0);
+        assertEq(tokenProxyV2.balanceOf(address(this)), 0);
```
MikeHathaway, line 181: Yeah, these values should still be 1 billion. They are 0 here because we attempted to extend our implementation with `ERC20PausableUpgradeable` which has a conflicting storage layout, so when the proxy looked in the slot it was storing `totalSupply()` it now finds one of the empty reserved slots from `ERC20PausableUpgradeable`

grandizzy: 👍 makes sense. I don't think we actually need the `ERC20PausableUpgradeable` at all?

src/Token.sol
grandizzy: Also here https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#what-does-it-mean-for-a-contract-to-be-upgrade-safe
Looks like constructor is not even called but initializer, where do you see constructor recommended by OZ?

test/Token.t.sol
MikeHathaway: Yeah I don't think we need `ERC20PausableUpgradeable` either. Had added it so that we can test the limits of the upgrades, and verify them programmatically. Will add it a separate `"implementationV3"` with expectations that there should be storage collisions

src/Token.sol
grandizzy: Ah, I see it here https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#initializing_the_implementation_contract

MikeHathaway: Yep, can also see it built out in tools they've produced like this: https://wizard.openzeppelin.com/#erc20

PR #6, opened 2022-08-18
Add onlyInitializing modifier for _mint function to prevent subsequen…

grandizzy:
…t mintings

Add SigUtils contract for create, hash, and sign the approvals off-chain
Implement additional tests:
- improve testRenounceOwnership test, check that upgrade and transferOwnership is no longer possible after renounceOwnership
- testCanMint by using a token contract which publicly expose mint function
- testTransferWithApprove and withPermit using fuzzed amounts (between 0 and max supply)
- vote related tests

TBD:
- is OK for testRemoveUpgradeability to be tested by renouncing to ownership?
- does testVote need to simulate an actual voting?

### DIFF
--- a/test/Token.t.sol
+++ b/test/Token.t.sol
@@ -5,6 +5,7 @@ import "forge-std/Test.sol";
 
 import { AjnaToken, UUPSProxy } from "../src/Token.sol";
 
+import { SigUtils }        from "./utils/SigUtils.sol";
 import { TestAjnaTokenV2 } from "./utils/TestAjnaTokens.sol";
 
 contract TokenTest is Test {
@@ -16,10 +17,14 @@ contract TokenTest is Test {
 
     AjnaToken internal _token;
     AjnaToken internal _tokenProxyV1;
+    SigUtils  internal _sigUtils;
     UUPSProxy proxy;
 
+    event Transfer(address indexed src, address indexed dst, uint256 wad);
     event Upgraded(address indexed implementation);
     event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
+    event DelegateChanged(address indexed delegator, address indexed fromDelegate, address indexed toDelegate);
+    event DelegateVotesChanged(address indexed delegate, uint256 previousBalance, uint256 newBalance);
 
     function setUp() external {
         _token = new AjnaToken();
@@ -30,6 +35,7 @@ contract TokenTest is Test {
         _tokenProxyV1 = AjnaToken(address(proxy));
 
         _tokenProxyV1.initialize();
+        _sigUtils = new SigUtils(_tokenProxyV1.DOMAIN_SEPARATOR());
     }
 
     function testCannotSendTokensToContract() external {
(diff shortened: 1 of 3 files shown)


test/Token.t.sol
```diff
@@ -177,7 +380,13 @@ contract TokenTest is Test {
     }
 
     function testRemoveUpgradeability() external {
-
+        // TODO: is OK to suppose that upgradeability can be removed by renouncing to ownership?
```
MikeHathaway, line 405: In the UUPS standard, like what we are using, the `upgradeTo` functionality is on the implementation itself. So theoretically upgrading should no longer be possible if ownership is renounced, but the `upgradeTo` method is still available if someone is able to find an exploit. I was imagining something for `testRemoveUpgradeability` along the lines of removing the upgrade method entirely, or making the upgrade method otherwise non-functional

test/Token.t.sol
```diff
@@ -83,46 +96,232 @@ contract TokenTest is Test {
...
+        _tokenProxyV1.transferFrom(address(this), address(5555), amount_);
+
+        assertEq(_tokenProxyV1.balanceOf(address(this)),  1_000_000_000 * 10 ** _tokenProxyV1.decimals() - amount_);
+        assertEq(_tokenProxyV1.balanceOf(address(5555)), amount_);
     }
 
-    // TODO: implement this
-    function testTransferWithPermit() external {
-        assertTrue(true);
+    function testTransferWithPermit(uint256 amount_) external {
```
MikeHathaway, line 131: We should potentially add a `transferWithPermit` method to the token contract, that takes in the permit signature, verifies it with `ecrecover` and then executes the transfer method so that a user can go from 0 allowance, to transferring the token in one tx as opposed to two. We have an example of this already with our LP NFTs: https://github.com/ajna-finance/contracts/blob/diez/src/base/PermitERC721.sol#L118

That being said, `permit` is still useful as it would enable us to implement things like `voteWithPermit` in the ecosystem fund contract taking in the offchain signature components and avoiding the use of multiple TX.

test/Token.t.sol
```diff
@@ -83,46 +96,232 @@ contract TokenTest is Test {
...
+        assertEq(_tokenProxyV1.getPastVotes(delegate1, 11111), 1_000 * 10 **18);
+        assertEq(_tokenProxyV1.getPastVotes(delegate2, 11111), 0);
+        assertEq(_tokenProxyV1.getPastVotes(delegate1, 11112), 0);
+        assertEq(_tokenProxyV1.getPastVotes(delegate2, 11112), 1_000 * 10 **18);
     }
 
     // TODO: implement this
     function testVote() external {
-
+        // TODO: should we implement this by actually simulate voting?
```
MikeHathaway: Yeah this is a good question - probably not worth it at this point since we would have to deploy something like https://docs.openzeppelin.com/contracts/4.x/governance#governor which would likely actually be the contract to implement `vote`. This is probably too heavy of a lift for the rewards. This work is also probably not fungible to the final design.

MikeHathaway (approved):
LGTM! Thanks for building out the tests

test/Token.t.sol
grandizzy: then probably we can just `revert("Not upgradeable");` in `_authorizeUpgrade` function?

test/Token.t.sol
grandizzy: added with https://github.com/ajna-finance/ajna-token/pull/6/commits/250cc5f04e31fcc4a3f4ae59775f361cf344650c

test/Token.t.sol
grandizzy: yep, removed for now with https://github.com/ajna-finance/ajna-token/pull/6/commits/250cc5f04e31fcc4a3f4ae59775f361cf344650c

PR #7, opened 2022-08-19
Implement a potential token wrapper design

MikeHathaway:


MikeHathaway:
We could also implement Burn, Permit, and Vote in the base token contract and then use the wrapper contract as a fallback if needed.

grandizzy (approved):
Looks good! I added a PR on top of it with additional tests like voting tests in Wrapped token https://github.com/ajna-finance/ajna-token/pull/8

grandizzy:
> 

I am in favor of keeping the base token as simpler as we can and add new functionalities in wrapper

PR #11, opened 2022-08-21
Pass address to mint tokens as constructor argument such as when depl…

grandizzy:
…oying tokens can be minted to another address

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ out/
 !/broadcast
 /broadcast/*
 /broadcast/*/31337/
+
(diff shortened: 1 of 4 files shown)


MikeHathaway (approved):
LGTM! Let's add `ERC20Votes` to `AjnaToken` as well - avoid usage of wrapper unless necessary

grandizzy:
> LGTM! Let's add `ERC20Votes` to `AjnaToken` as well - avoid usage of wrapper unless necessary

yep, in the process of porting this and tests

PR #12, opened 2022-08-22
Make base AjnaToken ERC20Votes, port voting related tests

grandizzy:


MikeHathaway (approved):
LGTM!

MikeHathaway:
After thinking about it some more - should we remove the before TokenTransfer check entirely? We are raising the cost of all token transfers for all users in order to protect a small subset of users from making an uncommon mistake

grandizzy:
> After thinking about it some more - should we remove the before TokenTransfer check entirely? We are raising the cost of all token transfers for all users in order to protect a small subset of users from making an uncommon mistake

I am fine with that and can do the change, I'd like first to hear others opinion too @ith-harvey @EdNoepel

EdNoepel:
> > After thinking about it some more - should we remove the before TokenTransfer check entirely? We are raising the cost of all token transfers for all users in order to protect a small subset of users from making an uncommon mistake
> 
> I am fine with that and can do the change, I'd like first to hear others opinion too @ith-harvey @EdNoepel

How much does it raise the cost of transfer?

grandizzy:
> > > After thinking about it some more - should we remove the before TokenTransfer check entirely? We are raising the cost of all token transfers for all users in order to protect a small subset of users from making an uncommon mistake
> > 
> > 
> > I am fine with that and can do the change, I'd like first to hear others opinion too @ith-harvey @EdNoepel
> 
> How much does it raise the cost of transfer?

It is 29891 with protection vs 29812 without

EdNoepel:
> > > > After thinking about it some more - should we remove the before TokenTransfer check entirely? We are raising the cost of all token transfers for all users in order to protect a small subset of users from making an uncommon mistake
> > > 
> > > 
> > > I am fine with that and can do the change, I'd like first to hear others opinion too @ith-harvey @EdNoepel
> > 
> > 
> > How much does it raise the cost of transfer?
> 
> It is 29891 with protection vs 29812 without

That's less than $0.02 of ETH; recommend leaving it in.

grandizzy:
I am going to merge this as it for now and prepare the other PRs on top of it. then we can remove `_beforeTokenTransfer` function in a subsequent PR if team decides so

PR #13, opened 2022-08-22
Ci slither

grandizzy:
- added GH actions slither analyzer and config file
- excluded `solc-version` detector. If using it it complains about solc `0.8.14` which is too recent to be trusted
```
Pragma version0.8.14 (src/BaseToken.sol#2) necessitates a version too recent to be trusted. Consider deploying with 0.6.12/0.7.6/0.8.7
solc-0.8.14 is not recommended for deployment
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-versions-of-solidity
```
Another approach would be to not exclude informational either and to always pass analyzer target / check reports after

### DIFF
--- a/.github/workflows/slither.yml
+++ b/.github/workflows/slither.yml
@@ -1,0 +1,14 @@
+name: Slither Analysis
+on: [push]
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


MikeHathaway (approved):
LGTM

PR #18, opened 2022-08-23
cleaned up readme

ith-harvey:


MikeHathaway (approved):
Removal LGTM. Do we need that TODO in there?

ith-harvey:
> 

Thought it would be a good reminder... we don't need to have this done now but it should be done by the time we go to audit. I can remove it if you wish

PR #20, opened 2022-09-05
Implement the Standard Proposal Mechanism of the Growth Coordination Fund

MikeHathaway:
Initial implementation of Ajna Growth Coordination Fund Standard Proposal Mechanism, with associated tests. Inherits from OpenZeppeling Governor with the GovernorVotesQuorumFraction extension to enable compatibility with existing governance tooling such as tally.xyz.

<br>
*IMPLEMENTED*

- Override `propose()`, `_getVotes` and `_castVote` to ensure all calls to the growth fund governor logic flow through the screening -> funding process.
- Implement insertion-sort algorithm for determine top 10 vote receiving proposals in the one-stage-one-vote screening stage.
- Implementation of QV based funding stage with a `checkSlate` method used to verify off-chain solutions to the knapsack problem presented by the optimal set of proposal to fund.
- Override `execute()` to ensure only proposals that went through the Distribution cycle can be executed.
- Add structs and mappings to enable storage and access of information required to track proposals and distributions over time.
- Use OZ.checkpoints to allow `distributionId` to be retrieved based upon a given `block.number`

<br>
*TODO*

- Implement the Extraordinary Funding Mechanism
- Split GrowthFund.sol into separate abstract contracts for standard and extraordinary funding mechanisms.
- Implement Proposal state enum to maintain compatibility with Tally.xyz expectation: https://docs.tally.xyz/user-guides/tally-contract-compatibility/openzeppelin-governor
- Expand README for GrowthFund logic


src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+        // retrieve current distribution Id
+        uint256 currentDistributionId = getDistributionId();
+
+        // set the current block number as the checkpoint for the current block
+        (uint256 prevId, uint256 newId) = _distributionIdCheckpoints.push(currentDistributionId + 1);
+        return newId;
+    }
+
+    /**
+     * @notice Start a new Distribution Period and reset appropiate state.
```
EdNoepel: sp: appropriate

src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+
+        return super.propose(targets, values, calldatas, description);
+    }
+
+    /**
+     * @notice Execute a proposal that has been approved by the community.
+     * @dev    Calls out to Governor.execute()
+     * @return proposalId of the executed proposal.
+     */
+    function execute(address[] memory targets, uint256[] memory values, bytes[] memory calldatas, bytes32 descriptionHash) public payable override(Governor) returns (uint256) {
```
EdNoepel: The OZ superclass' `_afterExecute` method clears the proposal _after_ execution.  But do we need a reentrancy check to ensure the proposal cannot itself call `execute`?

src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+        hasScreened[account_] = true;
+
+        // vote for the given proposal
+        return super._castVote(proposal_.proposalId, account_, 1, "", "");
+    }
+
+    /**
+     * @notice Calculates the number of votes available to an account depending on the current stage of the Distribution Period.
+     * @dev    Overrides OpenZeppelin _getVotes implementation to ensure appropriate voting weight is always returned.
+     * @dev    Snapshot checks are built into this function to ensure accurate power is returned regardless of the caller.
```
EdNoepel: I'm having trouble understanding what _number of votes available_ means.  Superclass calls this the _voting weight_.  Perhaps adding a `@returns` natspec could clarify?

src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+     * @notice Restrict voter to only voting once during the screening stage.
+     * @dev    See {IGovernor-hasVoted}.
+     */
+    function hasVoted(uint256, address account_) public view override(IGovernor) returns (bool) {
+        if (hasScreened[account_]) return true;
+        else return false;
+    }
+
+    /**
+     * @notice Required ovveride; not currently used due to divergence in voting logic.
```
EdNoepel: sp: override

script/AjnaToken.s.sol
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.7;
+pragma solidity 0.8.16;
 
 import { Script } from "forge-std/Script.sol";
-import { AjnaToken } from "../src/BaseToken.sol";
+import { AjnaToken } from "../src/AjnaToken.sol";
```
ith-harvey, line 5: I'm a big fan of us sharing our design patterns... accross our products. I propose we match our `contracts` repo and simply perform the imports without specifying the contracts...

Thoughts?

src/interfaces/IGrowthFund.sol
```diff
@@ -0,0 +1,129 @@
+// SPDX-License-Identifier: MIT
+
+//slither-disable-next-line solc-version
+pragma solidity 0.8.16;
+
+/**
+ * @title Ajna Growth Coordination Fund
+ */
+interface IGrowthFund {
```
ith-harvey: Propose we change the name of the repository, respective contracts, variabes and methods it may apply to from `Growth Coordination` -> `Grant Coordination` as it was changed in the v10 spec

script/AjnaToken.s.sol
MikeHathaway: That sounds good to me. This was written before we switched the `contracts` repo over, but can easily change this as well.

src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+    /**
+     * @notice Calculates the number of votes available to an account depending on the current stage of the Distribution Period.
+     * @dev    Overrides OpenZeppelin _getVotes implementation to ensure appropriate voting weight is always returned.
+     * @dev    Snapshot checks are built into this function to ensure accurate power is returned regardless of the caller.
+     */
+    function _getVotes(address account_, uint256 blockNumber_, bytes memory stage_) internal view override(Governor, GovernorVotes) returns (uint256) {
+        QuarterlyDistribution memory currentDistribution = distributions[getDistributionId()];
+
+        // within screening period 1 token 1 vote
+        if (keccak256(stage_) == keccak256(bytes("Screening"))) {
```
EdNoepel: I assume string is better than an enum because the superclass takes bytes, and because it produces a better hash?

test/GrowthFundTestHelper.sol
```diff
@@ -0,0 +1,225 @@
...
+
+        changePrank(voter_);
+        vm.expectEmit(true, true, false, true);
+        emit VoteCast(voter_, proposalId_, support_, voteAllocatedEmit, "");
+        growthFund_.castVoteWithReasonAndParams(proposalId_, support_, reason, params);
+    }
+
+    // expects a list of Proposal structs
+    // filepath expected to be defined from root
+    function _loadProposalSlateJSON(string memory filePath) internal returns (IGrowthFund.Proposal[] memory) {
```
EdNoepel: From where is this called?

EdNoepel (commented):
No issues.

src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+    uint256 internal extraordinaryFundingBaseQuorum;
+
+    address public   votingTokenAddress;
+    AjnaToken  internal ajnaToken;
+
+    /**
+     * @notice Maximum percentage of tokens that can be distributed by the treasury in a quarter.
+     * @dev Stored as a Wad percentage.
+     */
+    uint256 public globalBudgetConstraint = Maths.wdiv(Maths.wad(2), Maths.wad(100));
```
ith-harvey: Shouldn't this be a `constant`?

src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+
+    /**
+     * @notice Length of the distribution period in blocks.
+     * @dev    Equivalent to the number of blocks in 90 days. Blocks come every 12 seconds.
+     */
+    uint256 public constant DISTRIBUTION_PERIOD_LENGTH = 648000; // 90 days
```
ith-harvey: `constant`?

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
```diff
@@ -0,0 +1,728 @@
...
+    function maximumQuarterlyDistribution() public view returns (uint256) {
+        uint256 growthFundBalance = ajnaToken.balanceOf(address(this));
+        return Maths.wmul(growthFundBalance, globalBudgetConstraint);
+    }
+
+    /**************************/
+    /*** Proposal Functions ***/
+    /**************************/
+
+    function propose(
```
ith-harvey: Where does the Natspec for this func live?

test/GrowthFundTestHelper.sol
MikeHathaway: This was originally part of a test that I removed. I was attempting to use Foundry's new `parseJson` functionality, but ran into issues decoding large `int` values. I decided to leave this helper function in so we could revisit fixture based testing following upgrades to foundry's json parser. Reference is here: https://book.getfoundry.sh/cheatcodes/parse-json?highlight=parseJson#parsejson

That being said, I'd be open to removing this function.

src/GrowthFund.sol
MikeHathaway: That's a good call out. I don't think reetrancy should be possible here since the `checkProposal` modifier looks at the function selector and restricts it to just `transfer()`, so any proposal that contained the calldata for reentering execute should fail at the proposal creation stage. That being said, I don't think it would hurt too much to add a nonreentrant modifier to this method, and I can add a test around reentrancy in case there's a way to hide the `execute()` calldata.

I added OpenZeppelin `ReentrancyGuard` to better protect the `execute()` method.

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: What's the question here?

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: I've updated the natspec accordingly. Let me know if this is more clear.

src/GrowthFund.sol
MikeHathaway: Yeah the bytes parameter is available and required here, so it enables me to avoid the additional computation of taking the `blockNumber` and using that to determine the appropriate enum value. Would also likely need to mix the two anyways in order to determine if the votes are part of the standard mechanism, or the extraordinary mechanism.

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    uint256 internal extraordinaryFundingBaseQuorum;
+
+    address public   votingTokenAddress;
+    AjnaToken  internal ajnaToken;
+
+    /**
+     * @notice Maximum percentage of tokens that can be distributed by the treasury in a quarter.
+     * @dev Stored as a Wad percentage.
+     */
+    uint256 public constant globalBudgetConstraint = 20000000000000000;
```
grandizzy: I suggest making constants internal

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    /*******************/
+    /*** Constructor ***/
+    /*******************/
+
+    constructor(IVotes token_)
+        Governor("AjnaEcosystemGrowthFund")
+        GovernorVotes(token_) // token that will be used for voting
+        GovernorVotesQuorumFraction(4) // percentage of total voting power required; updateable via governance proposal
+    {
+        extraordinaryFundingBaseQuorum = 50; // initialize base quorum percentrage required for extraordinary funding to 50%
```
grandizzy: shouldn't `extraordinaryFundingBaseQuorum`,  `votingTokenAddress` and `ajnaToken` be immutables instead setting them in constructor? Also I don't think we need both  `votingTokenAddress` and `ajnaToken` as  `votingTokenAddress` can be easily obtained if needed as `address(ajnaToken)`. Instead importing `import "./AjnaToken.sol";` we can just import the `IERC20` interface, no need to work with the implementation.

So instead `AjnaToken  internal ajnaToken;` we can have `address immutable ajnaAddress = 0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079` and use `IERC20(ajnaAddress).balanceOf` or so

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    /**
+     * @notice Retrieve the current QuarterlyDistribution distributionId.
+     */
+    function getDistributionId() public view returns (uint256) {
+        return _distributionIdCheckpoints.latest();
+    }
+
+    /**
+     * @notice Calculate the block at which the screening period of a distribution ends.
+     * @dev    Screening period is 80 days, funding period is 10 days. Total distributin is 90 days.
```
grandizzy: NIT typo `Total distribution is 90 days`

src/interfaces/IGrowthFund.sol
```diff
@@ -0,0 +1,124 @@
...
+    /***************/
+
+    /**
+     * @notice Contains proposals that made it through the screening process to the funding stage.
+     * @dev Mapping and uint array used for tracking proposals in the distribution as typed arrays (like Proposal[]) can't be nested.
+     */
+    struct QuarterlyDistribution {
+        uint256 id;                 // id of the current quarterly distribution
+        uint256 votesCast;          // total number of votes cast that quarter
+        uint256 startBlock;         // block number of the quarterly distrubtions start
```
grandizzy: NIT typo `// block number of the quarterly distributions start` same for next line

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    }
+
+    /**
+     * @notice Set a new DistributionPeriod Id.
+     * @dev    Increments the previous Id nonce by 1, and sets a checkpoint at the calling block.number.
+     * @return The new distribution period Id.
+     */
+    function _setNewDistributionId() private returns (uint256) {
+        // retrieve current distribution Id
+        uint256 currentDistributionId = getDistributionId();
```
grandizzy: recommend not to call external / public function here (as it requires more gas) but inline `_distributionIdCheckpoints.latest()`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        // set the current block number as the checkpoint for the current block
+        (uint256 prevId, uint256 newId) = _distributionIdCheckpoints.push(currentDistributionId + 1);
```
grandizzy: since `prevId` is not used this can be changed to `(, uint256 newId)` or better IMO with named params
`function _setNewDistributionId() private returns (uint256 newId_)` then `(, newId_) = _distributionIdCheckpoints.push(currentDistributionId + 1);` and get rid of return statement

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        if (block.number <= lastDistribution.endBlock) revert DistributionPeriodStillActive();
+
+        // set the distribution period to start at the current block
+        uint256 startBlock = block.number;
+        uint256 endBlock = startBlock + DISTRIBUTION_PERIOD_LENGTH;
+
+        // set new value for currentDistributionId
+        _setNewDistributionId();
+
+        uint256 newDistributionId = getDistributionId();
```
grandizzy: same as above, don't call external function here to get `newDistributionId`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        // create QuarterlyDistribution struct
+        QuarterlyDistribution storage newDistributionPeriod = distributions[newDistributionId];
+        newDistributionPeriod.id =  newDistributionId;
```
grandizzy: NIT equals alignments

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        newDistributionPeriod.startBlock = startBlock;
+        newDistributionPeriod.endBlock = endBlock;
+
+        emit QuarterlyDistributionStarted(newDistributionId, startBlock, endBlock);
+        return newDistributionPeriod.id;
```
grandizzy: gas cost is more efficient to use named params, e.g.

`function startNewDistributionPeriod() external returns (uint256 newDistributionId_) {`

then `newDistributionId_ = _distributionIdCheckpoints.latest();` and `newDistributionPeriod.id =  newDistributionId_;` and get rid of return statement

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        // check that there isn't currently an active distribution period
+        if (block.number <= lastDistribution.endBlock) revert DistributionPeriodStillActive();
+
+        // set the distribution period to start at the current block
+        uint256 startBlock = block.number;
+        uint256 endBlock = startBlock + DISTRIBUTION_PERIOD_LENGTH;
+
+        // set new value for currentDistributionId
+        _setNewDistributionId();
```
grandizzy: since we're reading and incrementing `currentDistributionId` already in `_setNewDistributionId()` I recommend changing signature and returning the new distribution id from `_setNewDistributionId` and delete next line / save gas (e.g. if using named return var from `startNewDistributionPeriod` function: `newDistributionId_ = _setNewDistributionId()`)

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    /**
+     * @notice Vote on a proposal in the funding stage of the Distribution Period.
+     * @dev    Votes can be allocated to multiple proposals, quadratically, for or against.
+     * @param  proposal_ The current proposal being voted upon.
+     * @param  account_  The voting account.
+     * @param  voter_    The voter data struct tracking available votes.
+     * @param  budgetAllocation_ The amount of votes being allocated to the proposal.
+     * @return The amount of votes allocated to the proposal.
+     */
+    function _fundingVote(Proposal storage proposal_, address account_, QuadraticVoter storage voter_, int256 budgetAllocation_) internal returns (uint256) {
```
grandizzy: `voter_.budgetRemaining` and `proposal_.qvBudgetAllocated` are both `int256` so we can get rid of `int256 allocationUsed = 0;` and use function argument `int256 budgetAllocation_`, e.g. this block can be rewritten as 
```
        voter_.budgetRemaining      += budgetAllocation_;
        proposal_.qvBudgetAllocated += budgetAllocation_;
```
then `emit VoteCast(account_, proposal_.proposalId, budgetAllocation_, "");` - I think is better to have `VoteCast` `weight` of type int256 and get rid of `support` too (as we'll see if it was against or in support by the sign of `weight` / `budgetAllocation_`)

No need to return something from this function, I'd say function name can be `_fundVote`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+            QuadraticVoter storage voter = quadraticVoters[currentDistribution.id][account_];
+
+            // this is the first time a voter has attempted to vote this period
+            if (voter.votingWeight == 0) {
+                voter.votingWeight = Maths.wpow(super._getVotes(account_, getScreeningPeriodEndBlock(currentDistribution) - 33, ""), 2);
+                voter.budgetRemaining = int256(voter.votingWeight);
+            }
+
+            // amount of quadratic budget to allocated to the proposal
+            int256 budgetAllocation = abi.decode(params_, (int256));
```
grandizzy: use named param so this becomes `budgetAllocation_ = abi.decode(params_, (int256));` and remove return statement (see also comment re renaming `_fundingVote` to `_funcVote` and removing returned value)

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        else {
+            allocationUsed = budgetAllocation_;
+
+            // update voter and proposal vote tracking
+            voter_.budgetRemaining -= allocationUsed;
+            proposal_.qvBudgetAllocated += allocationUsed;
+        }
+
+        // update proposal vote tracking in top ten array
+        uint256 distributionId = getDistributionId();
```
grandizzy: use internal call here to retrieve distribution id

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    /*************************/
+
+    // return the index of the proposalId in the array, else -1
+    function _findInArray(uint256 proposalId, Proposal[] storage array) internal view returns (int256 index) {
+        index = -1; // default value indicating proposalId not in the array
+
+        for (int i = 0; i < int(array.length);) {
+            //slither-disable-next-line incorrect-equality
+            if (array[uint256(i)].proposalId == proposalId) {
+                index = i;
```
grandizzy: since we found the index, shouldn't we break loop and exit here?

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    // TODO: move this into a sort library; investigate replacement with simple iterative sort
+    /*************************/
+    /*** Sorting Functions ***/
+    /*************************/
+
+    // return the index of the proposalId in the array, else -1
+    function _findInArray(uint256 proposalId, Proposal[] storage array) internal view returns (int256 index) {
+        index = -1; // default value indicating proposalId not in the array
+
+        for (int i = 0; i < int(array.length);) {
```
grandizzy: since `array` is of type storage `array.length` is costly as it will load storage on each iteration, we should `uint256 arrayLength = array.length` and then `for (int i = 0; i < int(arrayLength);) {` recommend explicit `int256` instead aliased `int`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        return super.quorum(blockNumber);
+    }
+
+    // TODO: move this into a sort library; investigate replacement with simple iterative sort
+    /*************************/
+    /*** Sorting Functions ***/
+    /*************************/
+
+    // return the index of the proposalId in the array, else -1
+    function _findInArray(uint256 proposalId, Proposal[] storage array) internal view returns (int256 index) {
```
grandizzy: let's use returned param naming `index_` instead `index`. I recommend changing function name to something more specific like instead `_findInArray` to `_findProposalIndex`?

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        // get pointers for comparing proposal slates
+        bytes32 currentSlateHash = currentDistribution.fundedSlateHash;
+        bytes32 newSlateHash = getSlateHash(fundedProposals_);
+
+        // check if current slate received more support than the current leading slate
+        if (currentSlateHash != 0) {
+            if (sum > _sumBudgetAllocated(fundedProposalSlates[distributionId_][currentSlateHash])) {
+
+                _updateFundedSlate(fundedProposals_, currentDistribution, newSlateHash);
```
grandizzy: instead passing the entire storage obj I suggest passing the current distribution id only and update the `fundedSlateHash` outside  `_updateFundedSlate`, this should save gas afaik

Also IMO this block can be rewritten with a named `newTopSlate_` variable as

```
        newTopSlate_ = currentSlateHash == 0  ||
            (currentSlateHash != 0 && (sum > _sumBudgetAllocated(fundedProposalSlates[distributionId_][currentSlateHash])));
        if (newTopSlate_) {
            _updateFundedSlate(fundedProposals_, currentDistribution_.id, newSlateHash);
            currentDistribution.fundedSlateHash = newSlateHash;
        }
```

and not have any kind of return statements. Since `_updateFundedSlate` is only used here we can remove it entirely and iterate inline / merge with for at line 252 somehow so we iterate only once the `fundedProposals_`?

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+     */
+    function checkSlate(Proposal[] calldata fundedProposals_, uint256 distributionId_) external returns (bool) {
+        QuarterlyDistribution storage currentDistribution = distributions[distributionId_];
+
+        // check that the function is being called within the challenge period
+        if (block.number <= currentDistribution.endBlock || block.number > currentDistribution.endBlock + 50400) {
+            return false;
+        }
+
+        uint256 gbc = maximumQuarterlyDistribution();
```
grandizzy: let's use an internal function here

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+     * @notice Check if a slate of proposals meets requirements, and maximizes votes. If so, update QuarterlyDistribution.
+     * @param  fundedProposals_ Array of proposals to check.
+     * @param  distributionId_ Id of the current quarterly distribution.
+     * @return Boolean indicating whether the new proposal slate was set as the new top slate for distribution.
+     */
+    function checkSlate(Proposal[] calldata fundedProposals_, uint256 distributionId_) external returns (bool) {
+        QuarterlyDistribution storage currentDistribution = distributions[distributionId_];
+
+        // check that the function is being called within the challenge period
+        if (block.number <= currentDistribution.endBlock || block.number > currentDistribution.endBlock + 50400) {
```
grandizzy: let's express `50400` as `14 hours` for better readability. which beg the question why 14 hours?

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+            }
+
+            unchecked {
+                ++i;
+            }
+        }
+
+        // get pointers for comparing proposal slates
+        bytes32 currentSlateHash = currentDistribution.fundedSlateHash;
+        bytes32 newSlateHash = getSlateHash(fundedProposals_);
```
grandizzy: better `keccak256(abi.encode(proposals_))` inline instead calling the external function `getSlateHash`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+     * @param  calldatas List of calldata to be executed. Should be the transfer() method.
+     * @return proposalId The id of the newly created proposal.
+     */
+    function propose(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        string memory description
+    ) public override(Governor) checkProposal(targets, values, calldatas) returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, keccak256(bytes(description)));
```
grandizzy: better inline `uint256(keccak256(abi.encode(targets, values, calldatas, descriptionHash)));` than calling external `hashProposal` function

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        //slither-disable-next-line assembly
+        assembly {
+            tokensRequested := mload(add(dataWithSig, 68))
+        }
+
+        // store new proposal information
+        Proposal storage newProposal = proposals[proposalId];
+        newProposal.proposalId = proposalId;
+        newProposal.distributionId = getDistributionId();
```
grandizzy: let's use `_distributionIdCheckpoints.latest()`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        return super.propose(targets, values, calldatas, description);
+    }
+
+    /**
+     * @notice Execute a proposal that has been approved by the community.
+     * @dev    Calls out to Governor.execute()
+     * @return proposalId of the executed proposal.
+     */
+    function execute(address[] memory targets, uint256[] memory values, bytes[] memory calldatas, bytes32 descriptionHash) public payable override(Governor) nonReentrant returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
```
grandizzy: better inline `uint256(keccak256(abi.encode(targets, values, calldatas, descriptionHash)));` than calling external `hashProposal` function. I suggest using named variable `proposalId_` as well and get rid of return statement

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        return super.propose(targets, values, calldatas, description);
+    }
+
+    /**
+     * @notice Execute a proposal that has been approved by the community.
+     * @dev    Calls out to Governor.execute()
+     * @return proposalId of the executed proposal.
+     */
+    function execute(address[] memory targets, uint256[] memory values, bytes[] memory calldatas, bytes32 descriptionHash) public payable override(Governor) nonReentrant returns (uint256) {
```
grandizzy: NIT function arguments should be named like `targets_`, `values_` etc.

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    /**
+     * @notice Submit a new proposal to the Growth Coordination Fund
+     * @dev    All proposals can be submitted by anyone. There can only be one value in each array. Interface inherits from OZ.propose().
+     * @param  targets List of contracts the proposal calldata will interact with. Should be the Ajna token contract for all proposals.
+     * @param  values List of values to be sent with the proposal calldata. Should be 0 for all proposals.
+     * @param  calldatas List of calldata to be executed. Should be the transfer() method.
+     * @return proposalId The id of the newly created proposal.
+     */
+    function propose(
+        address[] memory targets,
```
grandizzy: NIT function arguments should be named like `targets_`, `values_` etc.

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+     * @notice Execute a proposal that has been approved by the community.
+     * @dev    Calls out to Governor.execute()
+     * @return proposalId of the executed proposal.
+     */
+    function execute(address[] memory targets, uint256[] memory values, bytes[] memory calldatas, bytes32 descriptionHash) public payable override(Governor) nonReentrant returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
+        Proposal storage proposal = proposals[proposalId];
+
+        // check if propsal is in the fundedProposalSlates list
+        if (_findInArray(proposalId, fundedProposalSlates[proposal.distributionId][distributions[proposal.distributionId].fundedSlateHash]) == -1) {
```
grandizzy: I find these lines a little bit hard to read, I recommend 
```
uint256 distributionId = proposal.distributionId;
if (_findInArray(proposalId, fundedProposalSlates[distributionId][distributions[distributionId].fundedSlateHash]) == -1) {\
```
or something easier to read

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    // TODO: may want to replace the conditional checks of stage here with the DistributionPhase enum
+    /**
+     * @notice Vote on a proposal in the screening or funding stage of the Distribution Period.
+     * @dev Override channels all other castVote methods through here.
+     * @param proposalId_ The current proposal being voted upon.
+     * @param account_    The voting account.
+     * @param params_     The amount of votes being allocated in the funding stage.
+     */
+     function _castVote(uint256 proposalId_, address account_, uint8, string memory, bytes memory params_) internal override(Governor) returns (uint256) {
+        QuarterlyDistribution memory currentDistribution = distributions[getDistributionId()];
```
grandizzy: let's avoid calling external `getDistributionId()` here. Also down the code we're calling this again in `Proposal[] storage currentTopTenProposals = topTenProposals[getDistributionId()];` this can be reused

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        Proposal storage proposal = proposals[proposalId_];
+
+        uint256 screeningPeriodEndBlock = getScreeningPeriodEndBlock(currentDistribution);
```
grandizzy: let's `currentDistribution.endBlock - 20 hours` instead  calling external `getScreeningPeriodEndBlock` (change 72000 to hours in this function as well).

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        bytes memory stage;
+        uint256 votes;
+
+        // screening stage
+        if (block.number >= currentDistribution.startBlock && block.number <= screeningPeriodEndBlock) {
+            Proposal[] storage currentTopTenProposals = topTenProposals[getDistributionId()];
+            stage = bytes("Screening");
+            votes = _getVotes(account_, block.number, stage);
+
+            return _screeningVote(currentTopTenProposals, account_, proposal, votes);
```
grandizzy: since `_screeningVote` is called only here let's just pass the `distributionId` param instead `Proposal[] storage currentTopTenProposals` and load `topTenProposals[distributionId]` in `_screeningVote` body

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+                currentTopTenProposals_.pop();
+                currentTopTenProposals_.push(proposal_);
+
+                // sort top ten proposals
+                _insertionSortProposalsByVotes(currentTopTenProposals_);
+            }
+        }
+
+        // ensure proposal list is within expected bounds
+        require(topTenProposals[getDistributionId()].length <= 10 && topTenProposals[getDistributionId()].length > 0, "CV:LIST_MALFORMED");
```
grandizzy: use custom error here. We already have the `currentTopTenProposals_` loaded from storage, no need to reload it here (twice). see comments within `_castVote` function, we can pass `distributionId` from `_castVote` function (instead `Proposal[] storage currentTopTenProposals_`) and load from storage in this function. then reuse to check its length (and load length only once)

grandizzy: isn't there a way to ensure array is not > 10, that is to enforce a fixed size of 10?

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        // funding stage
+        else if (block.number > screeningPeriodEndBlock && block.number <= currentDistribution.endBlock) {
+            stage = bytes("Funding");
+
+            QuadraticVoter storage voter = quadraticVoters[currentDistribution.id][account_];
+
+            // this is the first time a voter has attempted to vote this period
+            if (voter.votingWeight == 0) {
+                voter.votingWeight = Maths.wpow(super._getVotes(account_, getScreeningPeriodEndBlock(currentDistribution) - 33, ""), 2);
```
grandizzy: let's `currentDistribution.endBlock - 20 hours` instead  calling external `getScreeningPeriodEndBlock`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        if (keccak256(stage_) == keccak256(bytes("Screening"))) {
+            // calculate voting weight based on the number of tokens held before the start of the distribution period
+            return currentDistribution.startBlock == 0 ? 0 : super._getVotes(account_, currentDistribution.startBlock - 33, "");
+        }
+        // else if in funding period quadratic formula squares the number of votes
+        else if (keccak256(stage_) == keccak256(bytes("Funding"))) {
+            QuadraticVoter memory voter = quadraticVoters[currentDistribution.id][account_];
+            // this is the first time a voter has attempted to vote this period
+            if (voter.votingWeight == 0) {
+                return Maths.wpow(super._getVotes(account_, getScreeningPeriodEndBlock(currentDistribution) - 33, ""), 2);
```
grandizzy: let's `currentDistribution.endBlock - 20 hours` instead  calling external `getScreeningPeriodEndBlock`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    /**
+     * @notice Vote on a proposal in the screening stage of the Distribution Period.
+     * @param currentTopTenProposals_ List of top ten vote receiving proposals that made it through the screening round.
+     * @param account_                The voting account.
+     * @param proposal_               The current proposal being voted upon.
+     * @param votes_                  The amount of votes being cast.
+     * @return                        The amount of votes cast.
+     */
+    function _screeningVote(Proposal[] storage currentTopTenProposals_, address account_, Proposal storage proposal_, uint256 votes_) internal returns (uint256) {
+        if (hasVoted(proposal_.proposalId, account_)) revert AlreadyVoted();
```
grandizzy: can be simplified to `if (hasScreened[account_]) revert AlreadyVoted();`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+        // update proposal votes counter
+        proposal_.votesReceived += votes_;
+
+        // check if proposal was already screened
+        int indexInArray = _findInArray(proposal_.proposalId, currentTopTenProposals_);
+
+        // check if the proposal should be added to the top ten list for the first time
+        if (currentTopTenProposals_.length < 10 && indexInArray == -1) {
```
grandizzy: read length only once and reuse it on line 496 where `else if(currentTopTenProposals_[currentTopTenProposals_.length - 1].votesReceived`

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+        // else if in funding period quadratic formula squares the number of votes
+        else if (keccak256(stage_) == keccak256(bytes("Funding"))) {
+            QuadraticVoter memory voter = quadraticVoters[currentDistribution.id][account_];
+            // this is the first time a voter has attempted to vote this period
+            if (voter.votingWeight == 0) {
+                return Maths.wpow(super._getVotes(account_, getScreeningPeriodEndBlock(currentDistribution) - 33, ""), 2);
+            }
+            // voter has already allocated some of their budget this period
+            else {
+                return uint256(voter.budgetRemaining);
```
grandizzy: I wonder if a budget can go negative or only to 0 and if we can keep it as uint256 instead int256 and do the casts all over the code?

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+    //slither-disable-next-line naming-convention
+    function COUNTING_MODE() public pure override(IGovernor) returns (string memory) {
+        return "support=bravo&quorum=for,abstain";
+    }
+
+    /**
+     * @notice Restrict voter to only voting once during the screening stage.
+     * @dev    See {IGovernor-hasVoted}.
+     */
+    function hasVoted(uint256, address account_) public view override(IGovernor) returns (bool) {
```
grandizzy: I suggest using named variable and simplifying here to `hasVoted = hasScreened[account_];` `return hasScreened[account_]` is fine too :)

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+
+pragma solidity 0.8.16;
+
+import "@oz/governance/Governor.sol";
+import "@oz/governance/extensions/GovernorVotes.sol";
+import "@oz/governance/extensions/GovernorVotesQuorumFraction.sol";
+import "@oz/governance/IGovernor.sol";
+import "@oz/governance/utils/IVotes.sol";
+import "@oz/security/ReentrancyGuard.sol";
+import "@oz/token/ERC20/IERC20.sol";
```
grandizzy: doesn't seem we need to import this one?

grandizzy (changes requested):
Nice work! Please see my comments, I am yet to understand OZ's governance contract, will get some more into these and come back with additional questions

src/GrowthFund.sol
```diff
@@ -0,0 +1,740 @@
...
+     * @notice Maximum percentage of tokens that can be distributed by the treasury in a quarter.
+     * @dev Stored as a Wad percentage.
+     */
+    uint256 public constant globalBudgetConstraint = 20000000000000000;
+
+    /**
+     * @notice Length of the distribution period in blocks.
+     * @dev    Equivalent to the number of blocks in 90 days. Blocks come every 12 seconds.
+     */
+    uint256 public constant DISTRIBUTION_PERIOD_LENGTH = 648000; // 90 days
```
grandizzy: better write it as `DISTRIBUTION_PERIOD_LENGTH = 90 days`

grandizzy (changes requested):

src/GrowthFund.sol
MikeHathaway: This doesn't work for block numbers, only timestamps

src/GrowthFund.sol
MikeHathaway: removed

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
grandizzy: 👍

src/interfaces/IGrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Removed in favor of using the return value from `_setNewDistributionId();`

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated to access from `proposal_.distributionId` instead

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: added `break`

src/GrowthFund.sol
MikeHathaway: used variable for array length and changed function signature from `storage` to `memory`. Also changed this to a `pure` function (both of which resulted in substantial gas savings).

Following those updates, using a variable for array length actually increases gas costs

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: `hours` also works only with timestamps. This is `50400 blocks` which is equivalent to 1 week given the newly consistent 12 seconds per block post merge. We offer a 1 week "Challenge" period where anyone can offer competing slates with a more optimal solution for funding distribution

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: I'm not sure what you're asking here - we could likely remove this `require` statement entirely as it should be enforced previously, but there's a hard requirement that <= 10 proposals make it through the screening stage.

src/GrowthFund.sol
MikeHathaway: Updated accordingly

src/GrowthFund.sol
MikeHathaway: I had left it as public to ensure that the UI can be sure to produce the same calculations as the contracts

src/GrowthFund.sol
MikeHathaway: Updated to access `distributionId` via `proposal.distributionId`

src/GrowthFund.sol
grandizzy: 👍

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
grandizzy: yep, sounds good if we can remove it or at least load length only once from storage. Maybe use custom error too

src/GrowthFund.sol
MikeHathaway: I agree that this would be a better approach for the `emit` but I had left in the extra parameters to ensure that we stayed compliant with third-party governance interfaces like tally.xyz: https://docs.tally.xyz/user-guides/tally-contract-compatibility/openzeppelin-governor

Updated to use `budgetAllocation_` in favor of local `allocationUsed` variable.

src/GrowthFund.sol
MikeHathaway: The return statement is required to maintain compatibility with `super._castVote()` interface. It's more gas efficient to use the return value from `_fundingVote`

grandizzy: 👍

src/GrowthFund.sol
MikeHathaway: Removed it entirely

MikeHathaway: Removed entirely.

src/GrowthFund.sol
MikeHathaway: It should only be able to go to 0, but converting it to uint256 results in additional casts in `fundingVote` and `castVote` to track votes allocated resulting in an actual increase in type casting

src/GrowthFund.sol
MikeHathaway: Updated accordingly and removed `updateFundedSlate` function

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Resolving this comment temporarily as it's causing my github page to glitch.

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: Updated

src/GrowthFund.sol
MikeHathaway: the `extraordinaryFundingBaseQuorum` might change in the future as EFM proposals pass but this could also be calculated in alternative ways - not clear yet how that will be implemented given EFM is a WIP.

Simplified to only use a single token address like recommended. Left it as mutable while development is ongoing in case we wanted to deploy to a testnet, for example. Working on an approach using `vm.etch` to hardcode it during testing, but running into issues.

src/GrowthFund.sol
```diff
@@ -0,0 +1,714 @@
...
+import "@oz/security/ReentrancyGuard.sol";
+import "@oz/token/ERC20/IERC20.sol";
+import "@oz/utils/Checkpoints.sol";
+
+import "./libraries/Maths.sol";
+
+import "./interfaces/IGrowthFund.sol";
+
+
+contract GrowthFund is IGrowthFund, Governor, GovernorVotesQuorumFraction, ReentrancyGuard {
```
grandizzy: I think we can remove `is Governor here` because `GovernorVotesQuorumFraction` is `GovernorVotes` which is `Governor`. Then `function _getVotes(address account_, uint256 blockNumber_, bytes memory stage_) internal view override(GovernorVotes)` and `function quorum(uint256 blockNumber) public view override(GovernorVotesQuorumFraction) returns (uint256) {`
Happy to leave it as it is if you think makes thing clearer

MikeHathaway: Think it would be better to leave the Governor inheritance in to be explicit about how this is working. That being said, might be able to remove the QuorumFraction inheritance with the upcoming EFM pr

grandizzy: yeah makes sense, closing this comment

src/GrowthFund.sol
```diff
@@ -0,0 +1,699 @@
...
+     * @return proposalId_ The id of the newly created proposal.
+     */
+    function propose(
+        address[] memory targets_,
+        uint256[] memory values_,
+        bytes[] memory calldatas_,
+        string memory description_
+    ) public override(Governor) checkProposal(targets_, values_, calldatas_) returns (uint256 proposalId_) {
+        // hash proposalId according to IGovernor-hashProposal
+        proposalId_ = uint256(keccak256(abi.encode(targets_, values_, calldatas_, keccak256(bytes(description_)))));
```
grandizzy: I think would be better to do first `proposalId_ = super.propose(targets_, values_, calldatas_, description_);` and then do our custom code. This will execute `Governor` logic first and avoid double computation of `proposalId_` + extra return line. At the same time, `Governor.propose` checks for targets, values and calldata lengths to be the same, so our `checkProposal` modifier can be changed to only check
```
if (targets_.length != 1) revert InvalidProposal();
```
instead
```
if (targets_.length != 1 || values_.length != 1 || calldatas_.length != 1) {
            revert InvalidProposal();
}
```

src/GrowthFund.sol
```diff
@@ -0,0 +1,699 @@
...
+     * @param calldatas_ List of calldatas to execute if the proposal is successful.
+     */
+    modifier checkProposal(address[] memory targets_, uint256[] memory values_, bytes[] memory calldatas_) {
+        // check proposal can only execute one calldata, with one target
+        if (targets_.length != 1 || values_.length != 1 || calldatas_.length != 1) {
+            revert InvalidProposal();
+        }
+
+        if (targets_[0] != ajnaTokenAddress) revert InvalidTarget();
+        if (values_[0] != 0) revert InvalidValues();
```
grandizzy: why should `values_[0]` be 0?
ah, is that because values is the amount of ETH to include and we're funding not with ETH but with Ajna tokens?

src/GrowthFund.sol
```diff
@@ -0,0 +1,699 @@
...
+        uint256 tokensRequested;
+
+        //slither-disable-next-line assembly
+        assembly {
+            tokensRequested := mload(add(dataWithSig, 68))
+        }
+
+        // store new proposal information
+        Proposal storage newProposal = proposals[proposalId_];
+        newProposal.proposalId = proposalId_;
```
grandizzy: Do we really need the `proposalId` to be recorded in `Proposal` struct? I see common pattern is to be computed based on `targets`, `values`, `calldatas` and `description` values or proposal Id passed in function as arguments, e.g.

in `_castVote` function we already have the param `uint256 proposalId_` which we can pass to `_screeningVote` and `_fundingVote` instead `Proposal storage proposal` and then passed further to `_findProposalIndex` function. Then proposal struct can be loaded from storage inside these functions instead at the very beginning of `_castVote` e.g. instead lines 357 and 358

```
        Proposal storage proposal = proposals[proposalId_];
        QuarterlyDistribution memory currentDistribution = distributions[proposal.distributionId];
```

we can only do 
```
QuarterlyDistribution memory currentDistribution = distributions[proposals[proposalId_].distributionId];
```

The `checkSlate` function can receive an array of funded proposal ids instead of `Proposal[]` (we should also add this to `IGrowthFund` interface as it is currently missing) and `getFundedProposalSlate` can retrieved list of funded proposal ids instead `Proposal[]` array (this one is also missing from `IGrowthFund` interface)

grandizzy (changes requested):

src/GrowthFund.sol
```diff
@@ -0,0 +1,699 @@
...
+     * @notice Vote on a proposal in the funding stage of the Distribution Period.
+     * @dev    Votes can be allocated to multiple proposals, quadratically, for or against.
+     * @param  proposal_ The current proposal being voted upon.
+     * @param  account_  The voting account.
+     * @param  voter_    The voter data struct tracking available votes.
+     * @param  budgetAllocation_ The amount of votes being allocated to the proposal.
+     * @return The amount of votes allocated to the proposal.
+     */
+    function _fundingVote(Proposal storage proposal_, address account_, QuadraticVoter storage voter_, int256 budgetAllocation_) internal returns (uint256) {
+        uint8  support = 1;
```
grandizzy: since `budgetAllocation_` is of type `int256` and `voter_.budgetRemaining` the same, I suggest we rewrite following block (406 to 419) as
```
        uint8  support = budgetAllocation_ < 0 ? 0 : 1;
        voter_.budgetRemaining += budgetAllocation_;
```
I suggest `budgetAllocated` to be named var `budgetAllocated_` in `_fundingVote` function signature and get rid of `return budgetAllocated;`

grandizzy (changes requested):

src/GrowthFund.sol
```diff
@@ -0,0 +1,699 @@
...
+        }
+        // voter is voting in support of the proposal
+        else {
+            // update voter budget remaining
+            voter_.budgetRemaining -= budgetAllocation_;
+        }
+
+        // update proposal vote tracking
+        proposal_.qvBudgetAllocated += budgetAllocation_;
+        topTenProposals[proposal_.distributionId][uint256(_findProposalIndex(proposal_.proposalId, topTenProposals[proposal_.distributionId]))].qvBudgetAllocated = proposal_.qvBudgetAllocated;
```
grandizzy: avoid loading `topTenProposals[proposal_.distributionId]` from storage twice here, we can load it once and then do something like `topTenProposals[uint256(_findProposalIndex(proposal_.proposalId, topTenProposals)].qvBudgetAllocated` which is more readable too. or have a different var `proposalIndex = uint256(_findProposalIndex(proposal_.proposalId, topTenProposals)` and then `topTenProposals[proposalIndex].qvBudgetAllocated = proposal_.qvBudgetAllocated`

grandizzy (changes requested):

src/GrowthFund.sol
```diff
@@ -0,0 +1,699 @@
...
+     * @dev    See {IGovernor-_countVote}.
+     */
+    function _countVote(uint256 proposalId, address account, uint8 support, uint256 weight, bytes memory) internal override(Governor) {}
+
+    /**
+     * @notice Required override used in Governor.state()
+     * @dev See {IGovernor-quorumReached}.
+     */
+    function _quorumReached(uint256) internal pure override(Governor) returns (bool) {
+        return true;
```
grandizzy: I think we should add a comment here why always return true and point to the `_voteSucceeded` as the only way we use to mark a proposal as `Succeeded` (similar with natspec for `votingDelay` where we have `Since no voting delay is implemented, this is hardcoded to 0`)

src/GrowthFund.sol
MikeHathaway: Yes exactly, that would be for an ETH transfer, and the `values` field is part of the standard interface so I left it and forced it to be 0.

grandizzy: 👍

src/GrowthFund.sol
MikeHathaway: Updated accordingly

src/GrowthFund.sol
MikeHathaway: If we removed `proposalId` from the `Proposal` struct than I don't think we would be able to execute the `_findProposalIndex` function? 

Also, using proposalIds instead of working with a calldata array would be much more expensive given the relative costs of working with calldata vs storage.

src/GrowthFund.sol
MikeHathaway: This won't work since voting affirmatively for a proposal would result in an increasing budget remaining. Can add a ternary for the `voter_.budgetRemaining` modification as well, but that results in a net increase in computation compared to the existing approach.

I don't see where there's a `budgetAllocated` param in the function signature? 

Also, what is the purpose of removing the return statement? A return statement is required in `_castVote` to comply with the interface, and duplicating the logic required to return budget allocation in uint terms results in increased gas as compared to just returning `_fundingVote`

src/GrowthFund.sol
MikeHathaway: Updated accordingly, definitely much more readable.

src/GrowthFund.sol
MikeHathaway: Good call, updated to use `proposalId = super.propose()`, as well as above modifier.

src/interfaces/IGrowthFund.sol
MikeHathaway: Renamed accordingly.

src/GrowthFund.sol
grandizzy: makes sense, I missed this

src/GrowthFund.sol
grandizzy: > This won't work since voting affirmatively for a proposal would result in an increasing budget remaining. Can add a ternary for the `voter_.budgetRemaining` modification as well, but that results in a net increase in computation compared to the existing approach.
> 
> I don't see where there's a `budgetAllocated` param in the function signature?

Sorry, not `budgetAllocated` but `budgetAllocation_` I mean this one https://github.com/ajna-finance/GrowthCoordinationFund/blob/growth-fund-proto/src/GrantFund.sol#L391 which then we check if lower than 0 and add to `budgetRemaining`: `voter_.budgetRemaining += budgetAllocation_` else subtract from `budgetRemaining`
Doesn't seem right to me since it's an `int256` type we can just use it's sign and avoid checking if negative... Was wrong in initial comment  I think it should be `voter_.budgetRemaining -= budgetAllocation_` instead `+=`
> 
> Also, what is the purpose of removing the return statement? A return statement is required in `_castVote` to comply with the interface, and duplicating the logic required to return budget allocation in uint terms results in increased gas as compared to just returning `_fundingVote`

My suggestion is to use named returned variable (not get rid of return entirely) as afaik is better o  gas costs, that instead declaring it here https://github.com/ajna-finance/GrowthCoordinationFund/blob/growth-fund-proto/src/GrantFund.sol#L420 and returning it here https://github.com/ajna-finance/GrowthCoordinationFund/blob/growth-fund-proto/src/GrantFund.sol#L422 you could just have `_fundingVote returns (uint256 budgetAllocated_)` and set value of `budgetAllocated_` at line 420

MikeHathaway: Ah I see, I misunderstood the return variable point. Updated to use named return variable. 

Might be missing something, but flipping the sign still doesn't work for budgetRemaining, it just results in a vote against increasing their voting budget.

grandizzy (approved):
all good, let's merge it and we can follow up on other points in separate PRs

PR #22, opened 2022-11-01
Extraordinary Funding Mechanism

MikeHathaway:
- Implement Extraordinary Funding Mechanism
- Override OpenZeppelin.Governor.propose and OZ.Governor.state()
- Use our own Proposal structs to handle proposal state lifecycle
- Split each mechanism into separate abstract contracts that `GrantFund.sol` inherits from


src/GrantFund.sol
```diff
@@ -4,333 +4,99 @@ pragma solidity 0.8.16;
...
-    function _sumBudgetAllocated(Proposal[] memory proposalSubset_) internal pure returns (uint256 sum) {
-        sum = 0;
-        for (uint i = 0; i < proposalSubset_.length;) {
-            sum += uint256(proposalSubset_[i].qvBudgetAllocated);
-
-            unchecked {
-                ++i;
-            }
+            if (proposal.executed) return IGovernor.ProposalState.Executed;
+            else if (distribution.endBlock >= block.number && !voteSucceeded) return IGovernor.ProposalState.Active;
```
EdNoepel: Why do we need to check `voteSucceded` here?

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * @notice Execute an extraordinary funding proposal.
+     * @param targets_         The addresses of the contracts to call.
+     * @param values_          The amounts of ETH to send to each target.
+     * @param calldatas_       The calldata to send to each target.
+     * @param descriptionHash_ The hash of the proposal's description.
+     * @return proposalId_ The ID of the executed proposal.
+     */
+    function executeExtraordinary(address[] memory targets_, uint256[] memory values_, bytes[] memory calldatas_, bytes32 descriptionHash_) public nonReentrant returns (uint256 proposalId_) {
+        proposalId_ = hashProposal(targets_, values_, calldatas_, descriptionHash_);
```
EdNoepel, line 48: Should the proposal hash include the funding mechanism?

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,226 @@
...
+        // check proposal length is within limits of 1 month maximum and it hasn't already been submitted
+        if (block.number + MAX_EFM_PROPOSAL_LENGTH < endBlock_ || extraordinaryFundingProposals[proposalId_].proposalId != 0) {
+            revert ExtraordinaryFundingProposalInvalid();
+        }
+
+        uint256 totalTokensRequested = 0;
+
+        // check proposal attributes are valid
+        for (uint256 i = 0; i < targets_.length;) {
+
```
EdNoepel: Suggestion: Would be more readable to call a `_validateCallData` routine inside the loop.

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,226 @@
...
+            bytes4 selector;
+            //slither-disable-next-line assembly
+            assembly {
+                selector := mload(add(selDataWithSig, 0x20))
+            }
+            if (selector != bytes4(0xa9059cbb)) revert InvalidSignature();
+
+            // https://github.com/ethereum/solidity/issues/9439
+            // retrieve tokensRequested from incoming calldata, accounting for selector and recipient address
+            uint256 tokensRequested;
```
EdNoepel: Amount in the treasury can change between `proposeExtraordinary` and execution.  I agree `tokensRequested` is a better implementation, but it differs from requirements in section 9.2.2.4 of the whitepaper.

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        // check voting power at snapshot block
+        votes_ = _getVotes(account_, block.number, abi.encode(proposalId_));
+        proposal.votesReceived += votes_;
+
+        // check if the proposal has received more votes than minimumThreshold and tokensRequestedPercentage of all tokens
+        if (proposal.votesReceived >= proposal.tokensRequested + getPercentageOfTreasury(getMinimumThresholdPercentage())) {
+            proposal.succeeded = true;
+        }
```
EdNoepel: nit: expecting bat-else `} else {`

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,226 @@
...
+        else {
+            proposal.succeeded = false;
+        }
+
+        // record that voter has voted on this extraorindary funding proposal
+        hasScreened[proposalId_][account_] = true;
+
+        emit VoteCast(account_, proposalId_, 1, votes_, "");
```
EdNoepel, line 145: Note to reviewers: args found here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/IGovernor.sol#L55

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,226 @@
...
+        if (proposal.startBlock > block.number || proposal.endBlock < block.number || proposal.executed == true) {
+            revert ExtraordinaryFundingProposalInactive();
+        }
+
+        // check voting power at snapshot block
+        votes_ = _getVotes(account_, block.number, abi.encode(proposalId_));
+        proposal.votesReceived += votes_;
+
+        // check if the proposal has received more votes than minimumThreshold and tokensRequestedPercentage of all tokens
+        if (proposal.votesReceived >= proposal.tokensRequested + getPercentageOfTreasury(getMinimumThresholdPercentage())) {
```
EdNoepel: Let's say there's 1,000,000 tokens in treasury and it's the second vote.  Proposer requests 200,000 tokens (20%).  This means 50+20+5=75%, or 750,000 tokens should be required.

This logic would take 200,000 + (0.55 * 1,000,000) = 750,000.  So yeah, looks good.

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,226 @@
...
+            return 0.500000000000000000 * 1e18 + (fundedExtraordinaryProposals.length * (0.050000000000000000 * 1e18));
+        }
+    }
+
+    /**
+     * @notice Get the number of ajna tokens equivalent to a given percentage.
+     * @param percentage_ The percentage of the treasury to retrieve, in WAD.
+     * @return The number of tokens, in WAD.
+     */
+    function getPercentageOfTreasury(uint256 percentage_) public view returns (uint256) {
```
EdNoepel: Method name is confusing, because it suggests the function will return a percentage.  But the natspec explains it nicely.

EdNoepel (commented):
Haven't reviewed the tests yet, but contracts look good upon first review.

src/base/ExtraordinaryFunding.sol
MikeHathaway: Agreed it could be better - what do you think about `getPercentageOfTreasuryTokenEquivalent`? More specific albeit Java-esque

src/base/ExtraordinaryFunding.sol
MikeHathaway: Updated

src/base/ExtraordinaryFunding.sol
MikeHathaway: Updated accordingly - the gas cost of using a common routine was extremely minimal

src/GrantFund.sol
MikeHathaway: We don't - I had added when maintaining parity between standard and extraordinary flows in `state()`. It's only needed in the Extraordinary mechanism, so I will remove it.

src/base/ExtraordinaryFunding.sol
MikeHathaway: This is a good call out. I will discuss it with business tomorrow with the purpose of updating the whitepaper to match this. I don't think it's possible to retroactively update the `tokensRequested` param in the calldata without altering the resulting `proposalId` hash, and it's likely easier to reason about a specified amount of tokens from a social layer perspective.

src/base/ExtraordinaryFunding.sol
EdNoepel: `getSliceOfTreasury`?  I feel like "percentage" shouldn't be in the name if it doesn't return a percentage.

MikeHathaway: Sounds good, updated accordingly

src/base/ExtraordinaryFunding.sol
MikeHathaway: We could add a fifth parameter to it, but the `proposalId` generation process would then have to be different between Standard, and Extraordinary since Standard still utilizes the `Governor.execute()` method. That being said, they're both just hashes so shouldn't otherwise interfere. I had been thinking that we might want to create a requirement spec for the `description_` string, but it might be cleaner to just add the fifth parameter as `bytes("extraordinary")` and avoid any collisions that way.

EdNoepel: I'm happy using `description_` string to avoid collisions.

src/base/ExtraordinaryFunding.sol
MikeHathaway: The whitepaper has been updated to match the logic of the `tokensRequested` calldata based implementation.

EdNoepel (approved):
LGTM

src/GrantFund.sol
```diff
@@ -4,333 +4,99 @@ pragma solidity 0.8.16;
...
+     * @notice Given a proposalId, find if it is a standard or extraordinary proposal.
      */
-    function _setNewDistributionId() private returns (uint256 newId_) {
-        // retrieve current distribution Id
-        uint256 currentDistributionId = _distributionIdCheckpoints.latest();
-
-        // set the current block number as the checkpoint for the current block
-        (, newId_) = _distributionIdCheckpoints.push(currentDistributionId + 1);
+    function findMechanismOfProposal(uint256 proposalId_) public view returns (uint8) {
+        if (standardFundingProposals[proposalId_].proposalId != 0) return 0; // 0 = standard funding proposal
```
grandizzy: recommend using enums for better clarity, e.g. `enum FundingProposalMechanism{ STANDARD, EXTRAORDINARY }`

src/GrantFund.sol
```diff
@@ -4,333 +4,99 @@ pragma solidity 0.8.16;
...
-    function _sumBudgetAllocated(Proposal[] memory proposalSubset_) internal pure returns (uint256 sum) {
-        sum = 0;
-        for (uint i = 0; i < proposalSubset_.length;) {
-            sum += uint256(proposalSubset_[i].qvBudgetAllocated);
-
-            unchecked {
-                ++i;
-            }
+            if (proposal.executed) return IGovernor.ProposalState.Executed;
+            else if (distribution.endBlock >= block.number) return IGovernor.ProposalState.Active;
```
grandizzy: in extraordinary check we do `else if (proposal.endBlock >= block.number && !voteSucceeded) return IGovernor.ProposalState.Active;` shouldn't `voteSucceeded` be checked here as well?

src/GrantFund.sol
```diff
@@ -4,333 +4,99 @@ pragma solidity 0.8.16;
...
-                tokensRequested := mload(add(tokenDataWithSig, 68))
-            }
-
-            // update tokens requested for additional calldata
-            newProposal.tokensRequested += tokensRequested;
-
-            unchecked {
-                ++i;
-            }
+            if (proposal.executed) return IGovernor.ProposalState.Executed;
```
grandizzy: I think we can externalize this in a  `_getProposalState` function and reuse code, that is instead having 2 blocks of code to determine proposal state we could have only one. also not loading entire structs in memory could save some gas, so `state` function can look something like

```

    function state(uint256 proposalId_) public view override(Governor) returns (IGovernor.ProposalState) {
        uint8 mechanism = findMechanismOfProposal(proposalId_);

        // standard proposal state checks
        if (mechanism == FundingProposalMechanism.STANDARD) {
            return _getProposalState(
                standardFundingProposals[proposalId_].executed,
                distributions[_distributionIdCheckpoints.latest()].endBlock,
                _standardFundingVoteSucceeded(proposalId_)
            );
        }
        // extraordinary funding proposal state checks
        else if (mechanism == FundingProposalMechanism.EXTRAORDINARY) {
            return _getProposalState(
                extraordinaryFundingProposals[proposalId_].executed,
                extraordinaryFundingProposals[proposalId_].endBlock,
                _extraordinaryFundingVoteSucceeded(proposalId_)
            );
        }
    }

    function _getProposalState(bool executed, uint256 endBlock, bool voteSucceeded) internal view returns (IGovernor.ProposalState) {
        if (executed) return IGovernor.ProposalState.Executed;
        else if (endBlock >= block.number && !voteSucceeded) return IGovernor.ProposalState.Active;
        else if (voteSucceeded) return IGovernor.ProposalState.Succeeded;
        else return IGovernor.ProposalState.Defeated;
    }
```

src/GrantFund.sol
```diff
@@ -4,333 +4,99 @@ pragma solidity 0.8.16;
...
+            else if (proposal.endBlock >= block.number && !voteSucceeded) return IGovernor.ProposalState.Active;
+            else if (voteSucceeded) return IGovernor.ProposalState.Succeeded;
+            else return IGovernor.ProposalState.Defeated;
         }
```
grandizzy, line 90: should we revert or return a `IGovernor.ProposalState.Defeated` if `mechanism` is not `STANDARD` nor `EXTRAORDINARY`?

src/base/StandardFunding.sol
```diff
@@ -0,0 +1,485 @@
...
+     * @dev distribution.id => proposalId => Proposal
+     */
+    mapping(uint256 => Proposal) standardFundingProposals;
+
+    /**
+     * @dev Mapping of distributionId to a sorted array of 10 proposals with the most votes in the screening period.
+     * @dev distribution.id => Proposal[]
+     * @dev A new array is created for each distribution period
+     */
+    mapping(uint256 => Proposal[]) topTenProposals;
```
grandizzy: if I correctly get this, we're going to keep all proposals in `standardFundingProposals` mapping: `mapping(uint256 => Proposal) standardFundingProposals;` (when executed we're not removing from mapping)
Do we really need here a mapping to the entire Proposal struct 
`mapping(uint256 => Proposal[]) topTenProposals`
or can this be simplified to
`mapping(uint256 => uint256[]) topTenProposals`
where we don't keep entire struct again but rather the `proposalId`s?
Then whenever we change something to the proposal we're only going to maintain the struct in `standardFundingProposals` and not update in 2 places (even 3 if we consider the `fundedProposalSlates` mapping bellow too)

src/base/StandardFunding.sol
```diff
@@ -0,0 +1,485 @@
...
+
+    /**
+     * @notice Mapping of quarterly distributions to a hash of a proposal slate to a list of funded proposals.
+     * @dev distributionId => slate hash => Proposal[]
+     */
+    mapping(uint256 => mapping(bytes32 => Proposal[])) fundedProposalSlates;
```
grandizzy: same as above, cannot this be simplified to `mapping(uint256 => mapping(bytes32 => uint256[])) fundedProposalSlates;` that is `distributionId => slate hash => proposalId[]` ?

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * @notice Mapping of extant extraordinary funding proposals.
+     * @dev proposalId => ExtraordinaryFundingProposal.
+     */
+    mapping (uint256 => ExtraordinaryFundingProposal) public extraordinaryFundingProposals;
+
+    /**
+     * @notice The list of extraordinary funding proposals that have been executed.
+     */
+    ExtraordinaryFundingProposal[] public fundedExtraordinaryProposals;
```
grandizzy: same comment as for the standard funding impl, cannot we just point here to `proposalId` rather than entire struct, so to have ` uint256[] public fundedExtraordinaryProposalsIds;` list of arrays and for other details to use struct within `mapping (uint256 => ExtraordinaryFundingProposal) public extraordinaryFundingProposals;` ?

grandizzy (changes requested):

src/GrantFund.sol
MikeHathaway: Updated

src/GrantFund.sol
MikeHathaway: I don't think `Defeated` should be used since the proposal should never become active or be open for consideration without passing the mechanism check, so any proposal that doesn't return a value is inherently an invalid argument.

src/GrantFund.sol
MikeHathaway: It's needed only in the extraordinary funding mechanism because proposals can be executed immediately upon passing the required threshold, whereas in the standard flow, proposals can't be executed until the 1 week challenge period of providing off-chain calculated funding slates has completed. Until that week has passed, a proposal will be active but any success is inherently temporary and may change in the event of the provision of a superior slate, so active is likely less misleading.

src/GrantFund.sol
MikeHathaway: As explained above `voteSucceeded` is only a necessary check in the extraordinary flow, so the two sections logic would diverge. Avoiding loading the entire struct is a good call out, updated accordingly for significant gas savings.

src/base/ExtraordinaryFunding.sol
MikeHathaway: Updated accordingly

src/base/StandardFunding.sol
MikeHathaway: Good call out, updated accordingly for huge gas savings.

src/base/StandardFunding.sol
MikeHathaway: Updated accordingly

src/GrantFund.sol
```diff
@@ -476,40 +149,62 @@ contract GrantFund is IGrantFund, Governor, GovernorVotesQuorumFraction, Reentra
...
-            return super._getVotes(account_, blockNumber_, "");
-        }
-        // voting is not possible for non-specified pathways
         else {
-            return 0;
+            if (params_.length != 0) {
+                // attempt to decode a proposalId from the params
+                uint256 proposalId = abi.decode(params_, (uint256));
+
+                ExtraordinaryFundingProposal memory proposal = extraordinaryFundingProposals[proposalId];
```
grandizzy: I think we could save gas here too by not loading entire struct in memory (as we only need the `proposalId` and `startBlock`), just by
```
                // one token one vote for extraordinary funding
                if (extraordinaryFundingProposals[proposalId].proposalId != 0) {
                    uint256 proposalStartBlock = extraordinaryFundingProposals[proposalId].startBlock;
                    return _getVotesSinceSnapshot(account_, proposalStartBlock - 33, proposalStartBlock);
                }
```

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,191 @@
...
+    /***********************/
+
+    /**
+     * @notice Get the current minimum threshold percentage of Ajna tokens required for a proposal to exceed.
+     * @return The minimum threshold percentage, in WAD.
+     */
+    function getMinimumThresholdPercentage() public view returns (uint256) {
+        // default minimum threshold is 50
+        if (fundedExtraordinaryProposals.length == 0) {
+            return 0.500000000000000000 * 1e18;
```
grandizzy, line 163: probably not making a difference in terms of gas (as afaik Solidity will add this inline) but maybe we should make `0.500000000000000000 * 1e18` as a meaningful constant?

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,191 @@
...
+        }
+        // minimum threshold increases according to the number of funded EFM proposals
+        else {
+            return 0.500000000000000000 * 1e18 + (fundedExtraordinaryProposals.length * (0.050000000000000000 * 1e18));
```
grandizzy, line 167: should the second one be `0.05 * 1e18` or still `0.5 * 1e18`?

src/base/ExtraordinaryFunding.sol
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        uint256 totalTokensRequested = _validateCallDatas(targets_, values_, calldatas_);
+
+        // check tokens requested is within limits
+        if (totalTokensRequested > getSliceOfTreasury(Maths.WAD - getMinimumThresholdPercentage())) revert ExtraordinaryFundingProposalInvalid();
+
+        // store newly created proposal
+        ExtraordinaryFundingProposal storage newProposal = extraordinaryFundingProposals[proposalId_];
+        newProposal.proposalId = proposalId_;
```
grandizzy: NIT: would be nice to align these on `=` for better readability

src/base/StandardFunding.sol
```diff
@@ -0,0 +1,485 @@
...
+        // set the distribution period to start at the current block
+        uint256 startBlock = block.number;
+        uint256 endBlock = startBlock + DISTRIBUTION_PERIOD_LENGTH;
+
+        // set new value for currentDistributionId
+        newDistributionId_ = _setNewDistributionId();
+
+        // create QuarterlyDistribution struct
+        QuarterlyDistribution storage newDistributionPeriod = distributions[newDistributionId_];
+        newDistributionPeriod.id = newDistributionId_;
```
grandizzy: NIT: align these to `=` sign

src/base/StandardFunding.sol
```diff
@@ -0,0 +1,485 @@
...
+        (, newId_) = _distributionIdCheckpoints.push(currentDistributionId + 1);
+    }
+
+    /**
+     * @notice Start a new Distribution Period and reset appropriate state.
+     * @dev    Can be kicked off by anyone assuming a distribution period isn't already active.
+     * @return newDistributionId_ The new distribution period Id.
+     */
+    function startNewDistributionPeriod() external returns (uint256 newDistributionId_) {
+        QuarterlyDistribution memory lastDistribution = distributions[_distributionIdCheckpoints.latest()];
```
grandizzy: we could save some gas by not loading entire struct in memory and checking
```
if (block.number <= distributions[_distributionIdCheckpoints.latest()].endBlock) revert DistributionPeriodStillActive();
```
at line 123

src/base/StandardFunding.sol
```diff
@@ -0,0 +1,485 @@
...
+        }
+
+        return newTopSlate;
+    }
+
+    /**
+     * @notice Get the current maximum possible distribution of Ajna tokens that will be released from the treasury this quarter.
+     */
+    function maximumQuarterlyDistribution() public view returns (uint256) {
+        uint256 GrantFundBalance = IERC20(ajnaTokenAddress).balanceOf(address(this));
```
grandizzy: NIT: should be `uint256 grantFundBalance` with lowercase G

src/base/StandardFunding.sol
```diff
@@ -0,0 +1,485 @@
...
+
+        // check for duplicate proposals
+        if (standardFundingProposals[proposalId_].proposalId != 0) revert ProposalAlreadyExists();
+
+        // check params have matching lengths
+        if (targets_.length != values_.length || targets_.length != calldatas_.length || targets_.length == 0) revert InvalidProposal();
+
+        // store new proposal information
+        Proposal storage newProposal = standardFundingProposals[proposalId_];
+        newProposal.proposalId = proposalId_;
```
grandizzy: NIT: align these to `=` sign

src/base/StandardFunding.sol
```diff
@@ -0,0 +1,485 @@
...
+     * @dev Used to access information on the status of an ongoing distribution.
+     * @dev Updated at the start of each quarter.
+     */
+    Checkpoints.History internal _distributionIdCheckpoints;
+
+    /**
+     * @notice Mapping of quarterly distributions from the grant fund.
+     * @dev distributionId => QuarterlyDistribution
+     */
+    mapping(uint256 => QuarterlyDistribution) distributions;
```
grandizzy: I suggest specifying visibility of variables instead using the default

grandizzy (changes requested):
Looks good! only some more couple of comments / suggestions

src/GrantFund.sol
MikeHathaway: Updated accordingly

src/base/ExtraordinaryFunding.sol
MikeHathaway: It should be `0.05` as the threshold increases by 5% with each successful proposal

src/base/ExtraordinaryFunding.sol
MikeHathaway: Aligned

src/base/StandardFunding.sol
MikeHathaway: Aligned

src/base/StandardFunding.sol
MikeHathaway: Removed the variable entirely

src/base/StandardFunding.sol
MikeHathaway: Aligned

src/base/StandardFunding.sol
MikeHathaway: Updated accordingly

grandizzy (approved):
LGTM! Was also thinking at some other improvement (that we can follow up on subsequent PR) - since there's a `quarterly distribution` do we really need to keep distribution ids as uint256 data type or something like uint128 or uint64 should suffice?

src/base/StandardFunding.sol
MikeHathaway: Are you suggesting a change to the visibility here, or just make it explicit? The default would be `internal`, which I believe is the correct visibility since it's inherited, and getters are provided for external consumption

grandizzy: Ah, no change, only make it explicit

src/base/ExtraordinaryFunding.sol
MikeHathaway: Would normally agree with moving it out, but in this case that value is only used in that one function, and the threshold it's tied to also changes over time so thought it might be confusing to extract.

src/base/StandardFunding.sol
MikeHathaway: Updated accordingly

MikeHathaway:
> LGTM! Was also thinking at some other improvement (that we can follow up on subsequent PR) - since there's a `quarterly distribution` do we really need to keep distribution ids as uint256 data type or something like uint128 or uint64 should suffice?

So the `Checkpoints.History` utility natively operates upon `uint224`, reducing the bits might require casting

PR #23, opened 2022-11-11
add burn wrapper

MikeHathaway:


grandizzy:
can you refresh me why do we need this?

MikeHathaway:
> can you refresh me why do we need this?

The intention is to deploy one of these wrappers for each side-chain that we are intending to bridge to. That way, in the event of a bridge hack, we could protect the layer 1 and other chains from double spending of supposedly locked Ajna tokens.

grandizzy (approved):

PR #26, opened 2022-12-06
Ajna token deployment instructions

EdNoepel:
The one-liner seemed more appropriate for the README than a shell script.


MikeHathaway (approved):
LGTM!

PR #31, opened 2023-01-04
Fix Slither warnings and other cleanups

MikeHathaway:
- Fix all medium and high severity slither warnings. Only remaining slither warning are around solc version. We may want to downgrade to 0.8.7 across all contracts to match the token.
- Fix BurnWrapper bug to only allow wrapping AjnaTokens.
- Remove unused functions from Maths library.
- Styling clean ups
- Gas optimizations like loading from memory instead of storage.


src/base/StandardFunding.sol
```diff
@@ -23,7 +25,7 @@ abstract contract StandardFunding is Funding, IStandardFunding {
      * @notice Maximum percentage of tokens that can be distributed by the treasury in a quarter.
      * @dev Stored as a Wad percentage.
      */
-    uint256 internal constant globalBudgetConstraint = 20000000000000000;
+    uint256 internal constant GLOBAL_BUDGET_CONSTRAINT = .02 * 1e18;
```
ith-harvey: suggest to modify to `0.02 * 1e18`

ith-harvey (approved):
LGTM, admittedly not very knowledgeable about this code. Majority of changes seem to be return values and some slither cleanup.

src/BurnWrapper.sol
```diff
@@ -13,13 +13,30 @@ import { IERC20Metadata } from "@oz/token/ERC20/extensions/IERC20Metadata.sol";
 
 contract BurnWrappedAjna is ERC20, ERC20Burnable, ERC20Permit, ERC20Wrapper {
 
+    /**
+     * @notice Ethereum mainnet address of the Ajna Token.
+     */
+    address internal constant AJNA_TOKEN_ADDRESS = 0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079;
```
grandizzy: maybe make this immutable?

src/BurnWrapper.sol
```diff
@@ -13,13 +13,30 @@ import { IERC20Metadata } from "@oz/token/ERC20/extensions/IERC20Metadata.sol";
...
+     */
+    error InvalidWrappedToken();
+
     constructor(IERC20 wrappedToken)
         ERC20("Burn Wrapped AJNA", "bwAJNA")
         ERC20Permit("Burn Wrapped AJNA") // enables wrapped token to also use permit functionality
         ERC20Wrapper(wrappedToken)
-    {}
+    {
+        if (address(wrappedToken) != AJNA_TOKEN_ADDRESS) {
```
grandizzy, line 36: uh, nice catch!

src/base/ExtraordinaryFunding.sol
```diff
@@ -44,12 +44,12 @@ abstract contract ExtraordinaryFunding is Funding, IExtraordinaryFunding {
...
      * @return proposalId_ The ID of the executed proposal.
      */
-    function executeExtraordinary(address[] memory targets_, uint256[] memory values_, bytes[] memory calldatas_, bytes32 descriptionHash_) public nonReentrant returns (uint256 proposalId_) {
+    function executeExtraordinary(address[] memory targets_, uint256[] memory values_, bytes[] memory calldatas_, bytes32 descriptionHash_) external nonReentrant returns (uint256 proposalId_) {
         proposalId_ = hashProposal(targets_, values_, calldatas_, descriptionHash_);
 
         ExtraordinaryFundingProposal storage proposal = extraordinaryFundingProposals[proposalId_];
 
-        if (proposal.executed != false) {
+        if (proposal.executed) {
```
grandizzy, line 52: do I get this right that without this fix we are not able to execute  unexecuted proposals? wonder how the tests passed without this?

grandizzy (commented):
overall looks good, can you pls respond to the comments I put then we can merge

src/base/ExtraordinaryFunding.sol
MikeHathaway: This shouldn't change any behaviour, it just removes an unnecessary logic check. Instead of `if BOOL != false`, do `if BOOL`

src/base/StandardFunding.sol
MikeHathaway: done

src/BurnWrapper.sol
MikeHathaway: done

PR #32, opened 2023-01-06
Solc version

MikeHathaway:
- Downgrade AjnaToken contracts to match the deployed 0.8.7 bytecode. 
- Create two top level directories, `token` and `grants` to hold AjnaToken and GrantFund related logic respectively.
- Create a `MockAjnaToken` with the same solidity code at solc 0.8.16, and use that for the GrantFund tests.
- Update CI to slither check both top level directories, not just individual files. Also splits out code-coverage as a separate CI job.

TODO: 
- Use the deployable bytecode to deploy an 0.8.7 solc AjnaToken that uses a test address as the contract deployer and holder of initial funds. `vm.etch` that output, and call it during setup as an IERC20 token for the grant fund tests to ensure that the tests function as expected when interacting with `ERC20Votes` contracts that have been compiled with 0.8.7.
- Something roughly similar can be seen here: https://github.com/0xPhaze/paradigm-ctf-2021-foundry/blob/master/src/ParadigmSetup.sol


test/utils/MockAjnaToken.sol
```diff
@@ -0,0 +1,60 @@
...
+
+//slither-disable-next-line solc-version
+pragma solidity 0.8.16;
+
+import { ERC20 }         from "@oz/token/ERC20/ERC20.sol";
+import { ERC20Burnable } from "@oz/token/ERC20/extensions/ERC20Burnable.sol";
+import { ERC20Permit }   from "@oz/token/ERC20/extensions/draft-ERC20Permit.sol";
+import { ERC20Votes }    from "@oz/token/ERC20/extensions/ERC20Votes.sol";
+
+contract MockAjnaToken is ERC20, ERC20Burnable, ERC20Permit, ERC20Votes {
```
grandizzy: cannot we just extend here the `AjnaToken` instead copy / paste the code? Since MockAjnaToken is solidity 0.8.16 this should be having precedence over `AjnaToken` solidity version

grandizzy: nm, that's not the case, it barfs with `incompatible solidity versions`

grandizzy (approved):

prat-gpt (approved):
LGTM! just one update, added a PR which replaces `mockAjnaToken` with mainnet Ajna contract bytecode. Please have a look. https://github.com/ajna-finance/GrantCoordinationFund/pull/33

PR #38, opened 2023-01-31
Check duplicates s119

MikeHathaway:
- Add `_hasDuplicates` method to check if a proposalId is duplicated in a a slate being checked by `checkSlate`
- https://github.com/sherlock-audit/2023-01-ajna-judging/issues/119
- Fixes BurnWrapper tests that were failing due to difference in expected mainnet token supply in the fork. Since half of the mainnet tokens were burned, the assertions needed to use the new post burn balance and use a fork block after the burn block.

Before change gas:
```
| checkSlate                                  | 843             | 42408  | 56364  | 78425  | 12      |

```

After change gas (3 different runs):
```
| checkSlate                                  | 843             | 40259  | 51664  | 78999  | 13      |                                                                                                             
| checkSlate                                  | 843             | 41690  | 70263  | 78999  | 13      |                                                                                                             
| checkSlate                                  | 843             | 40259  | 51664  | 78999  | 13      |                                                                                                             
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,2 @@
 ## Ethereum node endpoint ##
-ETH_RPC_URL =
+ETH_RPC_URL=
--- a/src/grants/base/StandardFunding.sol
+++ b/src/grants/base/StandardFunding.sol
@@ -60,28 +60,28 @@ abstract contract StandardFunding is Funding, IStandardFunding {
     mapping(uint256 => uint256[]) internal topTenProposals;
 
     /**
-     * @notice Mapping of quarterly distributions to a hash of a proposal slate to a list of funded proposals.
-     * @dev distributionId => slate hash => proposalId[]
+     * @notice Mapping of a hash of a proposal slate to a list of funded proposals.
+     * @dev slate hash => proposalId[]
      */
-    mapping(uint256 => mapping(bytes32 => uint256[])) internal fundedProposalSlates;
+    mapping(bytes32 => uint256[]) internal fundedProposalSlates;
 
     /**
      * @notice Mapping of quarterly distributions to voters to a Quadratic Voter info struct.
      * @dev distributionId => voter address => QuadraticVoter 
      */
-    mapping (uint256 => mapping(address => QuadraticVoter)) internal quadraticVoters;
+    mapping(uint256 => mapping(address => QuadraticVoter)) internal quadraticVoters;
 
     /**
      * @notice Mapping of distributionId to whether surplus funds from distribution updated into treasury
      * @dev distributionId => bool
     */
-    mapping (uint256 => bool) internal isSurplusFundsUpdated;
+    mapping(uint256 => bool) internal isSurplusFundsUpdated;
 
     /**
      * @notice Mapping of distributionId to user address to whether user has claimed his delegate reward
      * @dev distributionId => address => bool
     */
-    mapping (uint256 => mapping (address => bool)) public hasClaimedReward;
+    mapping(uint256 => mapping (address => bool)) public hasClaimedReward;
 
     /*****************************************/
     /*** Distribution Management Functions ***/
@@ -111,7 +111,7 @@ abstract contract StandardFunding is Funding, IStandardFunding {
      */
     function _updateTreasury(uint256 distributionId_) private {
         QuarterlyDistribution memory currentDistribution =  distributions[distributionId_];
-        uint256[] memory fundingProposalIds = fundedProposalSlates[distributionId_][currentDistribution.fundedSlateHash];
+        uint256[] memory fundingProposalIds = fundedProposalSlates[currentDistribution.fundedSlateHash];
         uint256 totalTokensRequested;
         for (uint i = 0; i < fundingProposalIds.length; ) {
             Proposal memory proposal = standardFundingProposals[fundingProposalIds[i]];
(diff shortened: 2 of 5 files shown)


.env.example
```diff
@@ -1,2 +1,2 @@
 ## Ethereum node endpoint ##
-ETH_RPC_URL =
\ No newline at end of file
+ETH_RPC_URL=
```
ith-harvey, line 2: Don't believe this is a necessary change.

src/grants/base/StandardFunding.sol
```diff
@@ -191,6 +211,9 @@ abstract contract StandardFunding is Funding, IStandardFunding {
             return false;
         }
 
+        // check that the slate has no duplicates
+        if (_hasDuplicates(proposalIds_)) return false;
```
EdNoepel, line 212: Is this limited to 10 proposals (100 iterations)?  What is the gas impact of this check?

test/BurnWrappedToken.t.sol
```diff
@@ -16,14 +16,17 @@ contract BurnWrappedTokenTest is Test {
...
 
     event Transfer(address indexed from, address indexed to, uint256 value);
     event DelegateChanged(address indexed delegator, address indexed fromDelegate, address indexed toDelegate);
     event DelegateVotesChanged(address indexed delegate, uint256 previousBalance, uint256 newBalance);
 
     function setUp() external {
+        // create mainnet fork
         vm.createSelectFork(vm.envString("ETH_RPC_URL"));
+        // set fork block to block before token distribution
+        vm.rollFork(16527772);
```
EdNoepel, line 29: Burn happened in block 16478159, so this is after supply dropped to 1bbn.  All good.

EdNoepel (commented):
Concerned about gas implications of `checkSlate` changes.  Could someone DDOS by adding enough proposals to exceed block gas limit?

src/grants/base/StandardFunding.sol
MikeHathaway: Yes a slate is limited to at most 10 proposals. I'll add a before and after gas report

ith-harvey: +1 to this... Admittedly I don't know much about the growth coord code, Who would be the person calling this method?

Would it make any sense to have the actor adding to `proposalIds_` arr sort this array intentionally on insertion making it cheaper for the caller of `checkSlate` to check if there are duplicates?

test/BurnWrappedToken.t.sol
```diff
@@ -16,14 +16,17 @@ contract BurnWrappedTokenTest is Test {
     address internal _ajnaAddress   = 0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079; // mainnet ajna token address
     address internal _tokenDeployer = 0x666cf594fB18622e1ddB91468309a7E194ccb799; // mainnet token deployer
     address internal _tokenHolder   = makeAddr("_tokenHolder");
-    uint256 _initialAjnaTokenSupply = 2_000_000_000 * 1e18;
+    uint256 _initialAjnaTokenSupply = 1_000_000_000 * 1e18;
```
ith-harvey, line 19: Suggest a slightly longer explanation for this reason in the description.

.env.example
MikeHathaway: It doesn't work if spaces are present

EdNoepel: Agree with @MikeHathaway , example was incorrect.

src/grants/base/StandardFunding.sol
MikeHathaway: The caller is providing this array of proposals from off-chain indexing, so it should be entering pre sorted. Duplicate proposals would only enter from a malicious actor.

I've been investigating the usage of OpenZeppelin's EnnumerableSet library instead of the double for loop but the gas costs are 3x as high vs the simple iterative check (~75k gas). It's highly likely that the proposals being checked are < 10

MikeHathaway:
> Concerned about gas implications of `checkSlate` changes. Could someone DDOS by adding enough proposals to exceed block gas limit?

The gas impact is very minimal. Honest callers would be doing the work off-chain to provide a sorted list of proposals that have been funded and are optimal for maximizing the amount of votes < budget requirements. The maximum number of proposals is 10, but it's likely to be less e.g. if the optimal set of max votes < budget is only 3 proposals.

src/grants/base/StandardFunding.sol
EdNoepel: In the current implementation, I believe insert complexity is O(1), and validation complexity is O(n^2).

Ian's proposal is to validate the sort as passed in, such that a duplicate check would only need to compare neighbors.  That would give us insertion complexity of O(n) and validation complexity of O(n).  It would spread out computation costs, but after 10 insertions we're back to a total complexity of O(n^2 + n).

Probably not worth the hassle if this only costs 75k gas with 10 proposals.

EdNoepel (approved):

src/grants/base/StandardFunding.sol
MikeHathaway: I may not be understanding something, but I think in this context honest actors will be submitting slates via the SDK which will presort them into the optimal order. All incoming arrays are generated offchain. That proposal sounds like it would add extra gas costs for honest users which would already have a sorted array

ith-harvey: since it's only 10 I think my suggestion may be a bit of over-optimizing. LGTM as is if 10 or under : )

ith-harvey (approved):

dmitriia (commented):
Looks ok.

If all honest are presorted maybe just do a linear sort check, like it's done for ERC721Pool ids array, but with `>=` instead of `>`, and do it only here?

https://github.com/ajna-finance/contracts/blob/1cf13180e73bf6eb728dabc512447415f180aa93/src/ERC721PoolFactory.sol#L129-L141

```solidity
    /**
     *  @notice Check that the array of token ids is sorted in ascending order, else revert.
     *  @dev    The counters are modified in unchecked blocks due to being bounded by array length
     *  @param  tokenIds_ The array of token ids to check for sorting
     */
    function _checkTokenIdSortOrder(uint256[] memory tokenIds_) internal pure {
        for (uint256 i = 0; i < tokenIds_.length - 1; ) {
            if (tokenIds_[i] > tokenIds_[i + 1]) revert TokenIdsNotSorted();
            unchecked {
                ++i;
            }
        }
    }
```

PR #40, opened 2023-02-08
Sherlock 3

MikeHathaway:
- Prevent front running of `transferFromWithPermit` as mentioned in sherlock 3 https://github.com/sherlock-audit/2023-01-ajna-judging/issues/3 
- Someone could snipe the tx in the mempool, strip out the valid signature, and send a new transaction to the relaying spender with an updated to address and the valid signature
- Burn wrapper test fix is in [pr 38](https://github.com/ajna-finance/ecosystem-coordination/pull/38)

### DIFF
--- a/src/token/AjnaToken.sol
+++ b/src/token/AjnaToken.sol
@@ -44,7 +44,6 @@ contract AjnaToken is ERC20, ERC20Burnable, ERC20Permit, ERC20Votes {
      *  @notice Called by an owner of AJNA tokens to enable their tokens to be transferred by a spender address without making a seperate permit call
      *  @param  from_     The address of the current owner of the tokens
      *  @param  to_       The address of the new owner of the tokens
-     *  @param  spender_  The address of the third party who will execute the transaction involving an owners tokens
      *  @param  value_    The amount of tokens to transfer
      *  @param  deadline_ The unix timestamp by which the permit must be called
      *  @param  v_        Component of secp256k1 signature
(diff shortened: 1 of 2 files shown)


EdNoepel (approved):
The original Sherlock issue was excluded; thanks for digging it up.

grandizzy (approved):

dmitriia (commented):
Looks ok, but the change close no attack surfaces visible to me.

The spender _has to be_ `msg.sender` as otherwise transferFrom() will revert (or `msg.sender` will exchange allowance with `spender`, but that's strange usage without straightforward side-effects, the `msg.sender` can just do the direct transferFrom() in this case).

Furthermore, https://github.com/sherlock-audit/2023-01-ajna-judging/issues/3 isn't valid as `spender` is a part of the signature and if Kane inject himself as a spender the signature becomes incorrect:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Permit.sol#L49-L60

```solidity
    function permit(
        address owner,
        address spender,
        uint256 value,
        uint256 deadline,
        uint8 v,
        bytes32 r,
        bytes32 s
    ) public virtual override {
        require(block.timestamp <= deadline, "ERC20Permit: expired deadline");

        bytes32 structHash = keccak256(abi.encode(_PERMIT_TYPEHASH, owner, spender, value, _useNonce(owner), deadline));
```

I.e. if someone else be running it instead of spender they will need to spend their own allowance for transferFrom(). The signature is good only for `spender` that is encoded there, so sending it with another spender address will yield incorrect verification and be reverted. If Kane just changes `to`, tx will be reverted as he doesn't have the allowance for transferFrom().

The change itself is a good to go restriction, only `spender == msg.sender` is a valid usage anyway.

PR #43, opened 2023-02-09
Sherlock 64

MikeHathaway:
# Description of change
## High level

- Add `fundingVoteMulti` function to enable multiple funding votes to be cast in a single transaction.
- Add `FundingVoteParams` to track funding votes.
- Add `getFundingPowerVotes` which takes the sqrt of voting power to determine available discrete votes.
- split out delegate reward calculations and add `getDelegateRewards` external method to retrieve the delegate rewards available to a voter.

* Fix the funding stage voting process to use the sum of the squares of votes cast as highlighted in: https://github.com/sherlock-audit/2023-01-ajna-judging/issues/64

Regarding grant coordination voting of the funding stage, the technical spec states (page24):

Each address that holds Ajna tokens can vote as much as they want on every proposal, including negative votes, subject to the constraint that the sum of the squares of their votes cannot exceed the square of the number of Ajna tokens held.

The code enforces that the sum of all votes of a user (in absolute value) is lower or equal than the square of its token holdings. The votes are not squared before being summed.
Impact

This leads to a higher centralization/control risk than should be. Alice can deploy a smart contract that proposes a funding of all the available tokens for that round towards itself. The contract can reward people delegating to it in case of a successful funding.

Due to the incorrect tally, if 100 token holders each have 1 token, Alice only needs to convince 10 of them to join her cause to successfully gain the funds. She would receive 10^2 = 100 votes, while all other voters could only produce 90 * 1^2 = 90 votes.

If the technical spec were respected and with the same token distribution of 100 tokens split among 100 holders, Alice would need to bribe 50 token holders to join her cause to gain the funding. She has 50^2 voting power and can vote once with50 while all other can vote 50 * 1^2 = 50.

I consider this difference significant enough to be considered medium severity.

# Gas usage
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
```
| src/grants/GrantFund.sol:GrantFund contract |                 |        |        |        |         |
|---------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                             | Deployment Size |        |        |        |         |
| castVote                                    | 8223            | 86975  | 94747  | 111920 | 40      |
| castVoteWithReasonAndParams                 | 4758            | 116932 | 125977 | 166871 | 17      |
| checkSlate                                  | 908             | 51254  | 70256  | 148508 | 16      |
| claimDelegateReward                         | 1596            | 42208  | 58975  | 65775  | 13      |
| fundingVotesMulti                           | 159359          | 231516 | 174144 | 495660 | 6       |
```

### DIFF
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -128,17 +128,15 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
 
                 // this is the first time a voter has attempted to vote this period
                 if (voter.votingWeight == 0) {
-                    voter.votingWeight = Maths.wpow(_getVotesSinceSnapshot(account_, screeningPeriodEndBlock - 33, screeningPeriodEndBlock), 2);
-                    voter.budgetRemaining = int256(voter.votingWeight);
+                    // voter.votingWeight = Maths.wpow(_getVotesSinceSnapshot(account_, screeningPeriodEndBlock - 33, screeningPeriodEndBlock), 2);
+                    voter.votingWeight    = _getVotesSinceSnapshot(account_, screeningPeriodEndBlock - 33, screeningPeriodEndBlock);
+                    voter.budgetRemaining = Maths.wpow(voter.votingWeight, 2);
                 }
 
                 // amount of quadratic budget to allocated to the proposal
-                int256 budgetAllocation = abi.decode(params_, (int256));
+                int256 votes = abi.decode(params_, (int256));
 
-                // check if the voter has enough budget remaining to allocate to the proposal
-                if (Maths.abs(budgetAllocation) > voter.budgetRemaining) revert InsufficientBudget();
-
-                votesCast_ = _fundingVote(proposal, account_, voter, budgetAllocation);
+                votesCast_ = _fundingVote(proposal, account_, voter, votes);
             }
         }
 
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -105,9 +105,9 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
         uint256 screeningPeriodEndBlock = currentDistribution.endBlock - 72000;
 
         // this is the first time a voter has attempted to vote this period
-        if (voter.votingWeight == 0) {
-            voter.votingWeight    = _getVotesSinceSnapshot(msg.sender, screeningPeriodEndBlock - 33, screeningPeriodEndBlock);
-            voter.budgetRemaining = Maths.wpow(voter.votingWeight, 2);
+        if (voter.votingPower == 0) {
+            voter.votingPower    = Maths.wpow(_getVotesSinceSnapshot(msg.sender, screeningPeriodEndBlock - 33, screeningPeriodEndBlock), 2);
+            voter.remainingVotingPower = voter.votingPower;
         }
 
         for (uint256 i = 0; i < voteParams_.length; ) {
@@ -156,9 +156,9 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
                 QuadraticVoter storage voter = quadraticVoters[currentDistribution.id][account_];
 
                 // this is the first time a voter has attempted to vote this period
-                if (voter.votingWeight == 0) {
-                    voter.votingWeight    = _getVotesSinceSnapshot(account_, screeningPeriodEndBlock - 33, screeningPeriodEndBlock);
-                    voter.budgetRemaining = Maths.wpow(voter.votingWeight, 2);
+                if (voter.votingPower == 0) {
+                    voter.votingPower    = Maths.wpow(_getVotesSinceSnapshot(msg.sender, screeningPeriodEndBlock - 33, screeningPeriodEndBlock), 2);
+                    voter.remainingVotingPower = voter.votingPower;
                 }
 
                 // decode the amount of votes to allocated to the proposal
@@ -197,15 +197,13 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
         else if (keccak256(params_) == keccak256(bytes("Funding"))) {
             QuadraticVoter memory voter = quadraticVoters[currentDistribution.id][account_];
 
-            // TODO: fix voter available votes calculation
             // voter has already allocated some of their budget this period
-            if (voter.votingWeight != 0) {
-                // take the sqrt of the remaining budget to retrieve available votes
-                availableVotes_ = Maths.sqrt(voter.budgetRemaining);
+            if (voter.votingPower != 0) {
+                availableVotes_ = voter.remainingVotingPower;
             }
             // this is the first time a voter has attempted to vote this period
             else {
-                availableVotes_ = _getVotesSinceSnapshot(account_, currentDistribution.endBlock - 72033, currentDistribution.endBlock - 72000);
+                availableVotes_ = Maths.wpow(_getVotesSinceSnapshot(account_, currentDistribution.endBlock - 72033, currentDistribution.endBlock - 72000), 2);
             }
         }
         else {
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -112,7 +112,7 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
         // standard funding mechanism
         if (mechanism == FundingMechanism.Standard) {
             Proposal storage proposal = standardFundingProposals[proposalId_];
-            QuarterlyDistribution memory currentDistribution = distributions[proposal.distributionId];
+            QuarterlyDistribution storage currentDistribution = distributions[proposal.distributionId];
             uint256 screeningPeriodEndBlock = currentDistribution.endBlock - 72000;
 
             // screening stage
@@ -136,7 +136,7 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
                 int256 votes = abi.decode(params_, (int256));
 
                 // allocate the votes to the proposal
-                votesCast_ = _fundingVote(proposal, account_, voter, votes);
+                votesCast_ = _fundingVote(currentDistribution, proposal, account_, voter, votes);
             }
         }
 
@@ -167,9 +167,11 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
         else if (keccak256(params_) == keccak256(bytes("Funding"))) {
             QuadraticVoter memory voter = quadraticVoters[currentDistribution.id][account_];
 
+            // TODO: fix voter available votes calculation
             // voter has already allocated some of their budget this period
             if (voter.votingWeight != 0) {
-                availableVotes_ = voter.votingWeight - voter.votesUsed;
+                // take the sqrt of the remaining budget to retrieve available votes
+                availableVotes_ = Maths.sqrt(voter.budgetRemaining);
             }
             // this is the first time a voter has attempted to vote this period
             else {
@@ -239,7 +241,7 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
 
             // funding stage
             else if (block.number > screeningPeriodEndBlock && block.number <= currentDistribution.endBlock) {
-                hasVoted_ = quadraticVoters[currentDistribution.id][account_].votesUsed != 0;
+                hasVoted_ = quadraticVoters[currentDistribution.id][account_].votesCast.length != 0;
             }
         }
         else {
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -98,6 +98,35 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
     /*** Voting Functions ***/
     /************************/
 
+    function fundingVotesMulti(FundingVoteParams[] memory voteParams_) external returns (uint256 votesCast_) {
+        uint256 currentDistributionId = distributionIdCheckpoints.latest();
+        QuarterlyDistribution storage currentDistribution = distributions[currentDistributionId];
+        QuadraticVoter storage voter = quadraticVoters[currentDistribution.id][msg.sender];
+        uint256 screeningPeriodEndBlock = currentDistribution.endBlock - 72000;
+
+        // this is the first time a voter has attempted to vote this period
+        if (voter.votingWeight == 0) {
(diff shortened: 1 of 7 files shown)


src/grants/GrantFund.sol
```diff
@@ -127,18 +164,17 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
                 QuadraticVoter storage voter = quadraticVoters[currentDistribution.id][account_];
 
                 // this is the first time a voter has attempted to vote this period
-                if (voter.votingWeight == 0) {
-                    voter.votingWeight = Maths.wpow(_getVotesSinceSnapshot(account_, screeningPeriodEndBlock - 33, screeningPeriodEndBlock), 2);
-                    voter.budgetRemaining = int256(voter.votingWeight);
+                if (voter.votingPower == 0) {
+                    voter.votingPower          = Maths.wpow(_getVotesSinceSnapshot(msg.sender, screeningPeriodEndBlock - 33, screeningPeriodEndBlock), 2);
```
EdNoepel: I'm looking at this as `msg.sender` being the delegate, and `account_` the voter.  Why do we want to get the delegate's votes here instead of the voter?

src/grants/base/StandardFunding.sol
```diff
@@ -359,48 +378,80 @@ abstract contract StandardFunding is Funding, IStandardFunding {
...
-     * @param  budgetAllocation_ The amount of votes being allocated to the proposal.
-     * @return budgetAllocated_ The amount of votes allocated to the proposal.
+     * @notice Sum the square of each vote cast by a voter.
+     * @dev    Used to calculate if a voter has enough voting power to cast their votes.
+     * @param  votesCast_           The array of votes cast by a voter.
+     * @return votesCastSumSquared_ The sum of the square of each vote cast.
      */
-    function _fundingVote(Proposal storage proposal_, address account_, QuadraticVoter storage voter_, int256 budgetAllocation_) internal returns (uint256 budgetAllocated_) {
+    function _sumSquareOfVotesCast(FundingVoteParams[] memory votesCast_) internal pure returns (uint256 votesCastSumSquared_) {
+        uint256 numVotesCast = votesCast_.length;
```
EdNoepel, line 387: Is it practical to iterate through all votes here?  Why if there are thousands of voters?

src/grants/base/StandardFunding.sol
```diff
@@ -559,6 +626,29 @@ abstract contract StandardFunding is Funding, IStandardFunding {
...
+    /**
+     * @notice Identify where in an array of FundingVoteParams structs the proposal exists.
+     * @param proposalId_ The proposalId to search for.
+     * @param voteParams_ The array of FundingVoteParams structs to search.
+     * @return index_ The index of the proposalId in the array, else -1.
+     */
+    function _findProposalIndexOfVotesCast(uint256 proposalId_, FundingVoteParams[] memory voteParams_) internal pure returns (int256 index_) {
+        index_ = -1; // default value indicating proposalId not in the array
+
+        int256 numVotesCast = int256(voteParams_.length);
```
EdNoepel, line 646: Same concern as in `_sumSquareOfVotesCast`; unbounded iteration.

test/Maths.t.sol
```diff
@@ -53,6 +53,14 @@ contract MathsTest is Test {
 
     function testScaleConversions() external {
         assertEq(Maths.wad(153), 153 * 1e18);
-    } 
+    }
+
+    function testSqrt() external {
+        assertEq(Maths.sqrt(Maths.wad(100)) * 10**9, 10 * 1e18);
+        assertEq(Maths.sqrt(Maths.wad(100)) * 10**9, 10 * 1e18);
```
EdNoepel: nit: duplicate test line

src/grants/base/StandardFunding.sol
```diff
@@ -498,6 +557,11 @@ abstract contract StandardFunding is Funding, IStandardFunding {
         return fundedProposalSlates[slateHash_];
     }
 
+    /// @inheritdoc IStandardFunding
+    function getFundingPowerVotes(uint256 votingPower) external pure returns (uint256) {
+        return Maths.sqrt(votingPower) * 10**9;
```
EdNoepel: Why not incorporate the scaling factor into `Maths.sqrt`?

EdNoepel (commented):
Commented regarding some concerns.

src/grants/base/StandardFunding.sol
MikeHathaway: A voter can only vote on at most 10 proposals, so iteration is limited to <= 10. Each voter would call this separately

src/grants/base/StandardFunding.sol
MikeHathaway: Iteration is bounded by the number of screened proposals in the funding round being <= 10

test/Maths.t.sol
MikeHathaway: Will remove

src/grants/base/StandardFunding.sol
MikeHathaway: Updated accordingly

src/grants/base/StandardFunding.sol
EdNoepel: Understood.  👍

src/grants/base/StandardFunding.sol
```diff
@@ -359,48 +378,80 @@ abstract contract StandardFunding is Funding, IStandardFunding {
...
+
+        // check that the voter has enough budget remaining to cast the vote
+        if (quadraticTotalVotesUsed > voter_.votingPower) revert InsufficientBudget();
+
+        // update voter budget accumulator
+        voter_.remainingVotingPower = voter_.votingPower - quadraticTotalVotesUsed;
+
         // update total vote cast
-        currentDistribution.quadraticVotesCast += uint256(Maths.abs(budgetAllocation_));
+        currentDistribution_.fundingVotePowerCast += Maths.wpow(incrementalVotesUsed_, 2);
```
mattcushman: I still don't think this is correct, if I understand it.  incrementalVotedUsed here is the absolute value of the number of votes being cast.  But the vote power consumed by a vote isn't just the square of that.  Example:

Voter 1 votes 10 votes in favor of prop 1.  This would consume 10^2=100 votes.  All good there.

Voter 1 now votes 10 more votes in favor of prop 1.  The total cost is now (10+10)^2 = 400 votes, NOT just 10^2-=100 votes.

src/grants/base/StandardFunding.sol
mattcushman: where is this function used actually?

mattcushman (commented):
few comments... i find a lot of the variables/functions somewhat confusingly named.

src/grants/base/StandardFunding.sol
MikeHathaway: Good catch, I fixed the current distributions voting power accumulator `fundingVotePowerCast` in https://github.com/ajna-finance/ecosystem-coordination/pull/43/commits/abd5499e9df96bdb36c255132fd1f4646e863d26

The vote cost was already tracked properly on a per voter basis

src/grants/base/StandardFunding.sol
```diff
@@ -425,32 +428,40 @@ abstract contract StandardFunding is Funding, IStandardFunding {
...
         }
         // add the newly cast vote to the voter's votesCast array
         else {
             voter_.votesCast.push(voteParams_);
         }
 
-        // find the absolute number of incremental votes used
-        incrementalVotesUsed_ = uint256(Maths.abs(voteParams_.votesUsed));
+        // square the total votes on the proposal to find the incrementalVotingPowerUsed for this vote
+        uint256 incrementalVotingPowerUsed = Maths.wpow(uint256(Maths.abs(totalProposalVotesByVoter)), 2);
```
mattcushman: I think we need to remember the previous number of votes (existingVote.votesUsed) up above and set incrementalVotingPowerUsed = (existingVote.votesUsed + voteParams_.votesUsed)^2 - existingVote.votesUsed 

actually, I have some concerns about how this is going to work for delegators now that I think of it.  Its probably easier to explain on a call.. basically, if a bunch of people delegate their voting power, the delegator should get to vote all of those votes independently...

MikeHathaway: This comment looks to have been based upon old code. the approach was updated to something similar as suggested.

Due to the way delegation snapshotting occurs, we have no access to the per delegation tokens, only the total

mattcushman: i see, yes you're right, I missed the part where you decremented the voting power above

mattcushman (approved):
lgtm

src/grants/GrantFund.sol
MikeHathaway: the delegate and the voter are one and the same in all cases. People would need to explicitly delegate their tokens (either to themselves or others) prior to casting a vote. I updated this to to use `account_` to stay consistent with the other code blocks

src/grants/GrantFund.sol
```diff
@@ -98,6 +98,43 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
...
+
+            // cast each successive vote
+            votesCast_ += _fundingVote(
+                currentDistribution,
+                proposal,
+                msg.sender,
+                voter,
+                voteParams_[i]
+            );
+
```
ith-harvey, line 131: change to -> `unchecked {++i;}`

And provide comment about how this can't overflow

ith-harvey (approved):

dmitriia (commented):
This can malfunction as `fundingVotesReceived` can be negative, so abs looks to be needed:

https://github.com/ajna-finance/ecosystem-coordination/blob/1cf7daad2d229d3d220fab256da5fbdd56a47f84/src/grants/base/StandardFunding.sol#L165-L178

```diff
    /**
     * @notice Calculates the sum of funding votes allocated to a list of proposals.
     * @param  proposalIdSubset_ Array of proposal Ids to sum.
     * @return sum_ The sum of the funding votes across the given proposals.
     */
    function _sumProposalFundingVotes(uint256[] memory proposalIdSubset_) internal view returns (uint256 sum_) {
        for (uint i = 0; i < proposalIdSubset_.length;) {
-           sum_ += uint256(standardFundingProposals[proposalIdSubset_[i]].fundingVotesReceived);
+           sum_ += uint256(Maths.abs(standardFundingProposals[proposalIdSubset_[i]].fundingVotesReceived));

            unchecked {
                ++i;
            }
        }
    }
```

https://github.com/ajna-finance/ecosystem-coordination/blob/1cf7daad2d229d3d220fab256da5fbdd56a47f84/src/grants/base/StandardFunding.sol#L407-L454

```solidity
    function _fundingVote(QuarterlyDistribution storage currentDistribution_, Proposal storage proposal_, address account_, QuadraticVoter storage voter_, FundingVoteParams memory voteParams_) internal returns (uint256 incrementalVotesUsed_) {
        uint8  support = 1;
        uint256 proposalId = proposal_.proposalId;

        // determine if voter is voting for or against the proposal
        voteParams_.votesUsed < 0 ? support = 0 : support = 1;

        ...

        // update proposal vote tracking
        proposal_.fundingVotesReceived += voteParams_.votesUsed;
```

https://docs.soliditylang.org/en/latest/types.html#explicit-conversions

As I see that's not the change itself, it's a bug in master not spotted before:

https://github.com/sherlock-audit/2023-01-ajna/blob/main/ecosystem-coordination/src/grants/base/StandardFunding.sol#L177

MikeHathaway:
> This can malfunction as `fundingVotesReceived` can be negative, so abs looks to be needed:
> 
> https://github.com/ajna-finance/ecosystem-coordination/blob/1cf7daad2d229d3d220fab256da5fbdd56a47f84/src/grants/base/StandardFunding.sol#L165-L178
> 
> ```diff
>     /**
>      * @notice Calculates the sum of funding votes allocated to a list of proposals.
>      * @param  proposalIdSubset_ Array of proposal Ids to sum.
>      * @return sum_ The sum of the funding votes across the given proposals.
>      */
>     function _sumProposalFundingVotes(uint256[] memory proposalIdSubset_) internal view returns (uint256 sum_) {
>         for (uint i = 0; i < proposalIdSubset_.length;) {
> -           sum_ += uint256(standardFundingProposals[proposalIdSubset_[i]].fundingVotesReceived);
> +           sum_ += uint256(Maths.abs(standardFundingProposals[proposalIdSubset_[i]].fundingVotesReceived));
> 
>             unchecked {
>                 ++i;
>             }
>         }
>     }
> ```
> 
> https://github.com/ajna-finance/ecosystem-coordination/blob/1cf7daad2d229d3d220fab256da5fbdd56a47f84/src/grants/base/StandardFunding.sol#L407-L454
> 
> ```solidity
>     function _fundingVote(QuarterlyDistribution storage currentDistribution_, Proposal storage proposal_, address account_, QuadraticVoter storage voter_, FundingVoteParams memory voteParams_) internal returns (uint256 incrementalVotesUsed_) {
>         uint8  support = 1;
>         uint256 proposalId = proposal_.proposalId;
> 
>         // determine if voter is voting for or against the proposal
>         voteParams_.votesUsed < 0 ? support = 0 : support = 1;
> 
>         ...
> 
>         // update proposal vote tracking
>         proposal_.fundingVotesReceived += voteParams_.votesUsed;
> ```
> 
> https://docs.soliditylang.org/en/latest/types.html#explicit-conversions
> 
> As I see that's not the change itself, it's a bug in master not spotted before:
> 
> https://github.com/sherlock-audit/2023-01-ajna/blob/main/ecosystem-coordination/src/grants/base/StandardFunding.sol#L177

I believe it's not possible to reach this code with a negative int? If any of the proposals have a negative vote balance then it will return False out of `checkSlate` here: https://github.com/ajna-finance/ecosystem-coordination/blob/1cf7daad2d229d3d220fab256da5fbdd56a47f84/src/grants/base/StandardFunding.sol#L225. The `_sumProposalFundingVotes` is also only called from within `checkSlate`

dmitriia:
Yes, you are right. I think it will be the best to get rid of _sumProposalFundingVotes(), just record the sum and compare:

https://github.com/ajna-finance/ecosystem-coordination/blob/1cf7daad2d229d3d220fab256da5fbdd56a47f84/src/grants/base/StandardFunding.sol#L245-L247

```diff
        // check if slate of proposals is new top slate
        bool newTopSlate = currentSlateHash == 0 ||
-           (currentSlateHash!= 0 && sum > _sumProposalFundingVotes(fundedProposalSlates[currentSlateHash]))
+           (currentSlateHash!= 0 && sum > currentDistribution.totalProposalFundingVotes)
```

dmitriia (commented):
Looks ok

PR #44, opened 2023-02-14
Screening stage vote fix

MikeHathaway:
<!---
No need to add special tag
src/ & non src/ changes you need the following (that apply):
-->
# Description of change
## High level
* Enable voters to split their votes between multiple different proposals during the screening stage.
  * Removes the call to `super._castVote` in Governor as Governor presumes the usage of the full voting power on the proposal.
  * Add `screeningVotesCast` struct to track the number of votes cast on proposals in a given screening stage to prevent over voting. Use that struct in place of the previous `hasVotedScreening` struct.
  * Add `screeningVoteMulti` function similar to `fundingVoteMulti` which provides improved UX from voting on multiple screening stage proposals in one transaction.
* Fix bug that allowed people to vote outside of the correct distribution period or stage, with a new `InvalidVote` revert.
* Includes struct packing improvements https://github.com/ajna-finance/ecosystem-coordination/pull/47


# Gas usage
## Pre Change
```
| castVote                                    | 1406            | 34224  | 33609  | 59509  | 181     |
| castVoteWithReasonAndParams                 | 2185            | 81941  | 102577 | 167891 | 827     |
| checkSlate                                  | 906             | 45854  | 61706  | 79619  | 16      |
| claimDelegateReward                         | 1613            | 58362  | 58911  | 65711  | 397     |
| execute                                     | 2349            | 2349   | 2349   | 2349   | 1       |
| executeExtraordinary                        | 7125            | 81511  | 85437  | 141190 | 9       |
| executeStandard                             | 10841           | 60675  | 85263  | 95017  | 8       |
| fundingVotesMulti                           | 127669          | 217711 | 174084 | 497450 | 7       |
| proposeExtraordinary                        | 4468            | 78779  | 103720 | 110059 | 14      |
| proposeStandard                             | 4510            | 78694  | 83198  | 85198  | 75      |
| screeningVoteMulti                          | 3210            | 229206 | 162491 | 744884 | 5       |
| startNewDistributionPeriod                  | 1618            | 140313 | 152376 | 167334 | 18      |
```


## Post Change
```                                                                                                                                                                                                                  
| castVote                                    | 1402            | 33688  | 33738  | 35738  | 1017    |                                                                                                             
| castVoteWithReasonAndParams                 | 1911            | 75591  | 70317  | 125302 | 501     |                                                                                                             
| checkSlate                                  | 1028            | 45028  | 61002  | 78933  | 16      |
| claimDelegateReward                         | 1482            | 57897  | 58836  | 65636  | 234     |                                                                                                             
| execute                                     | 2457            | 2457   | 2457   | 2457   | 1       |                                                                                                             
| executeExtraordinary                        | 7343            | 55591  | 65511  | 121258 | 11      |                                                                                                             
| executeStandard                             | 9556            | 46292  | 63232  | 72987  | 8       |                                                                                                             
| findMechanismOfProposal                     | 614             | 2046   | 800    | 4726   | 3       |                                                                                                             
| fundingVotesMulti                           | 105054          | 182141 | 153813 | 396896 | 7       |                                                                                                                                                                                 
| propose                                     | 2922            | 2922   | 2922   | 2922   | 1       |                                                                                                             
| proposeExtraordinary                        | 4739            | 65677  | 82007  | 88346  | 16      |                                                                                                             
| proposeStandard                             | 4673            | 78099  | 82728  | 84728  | 61      |                                                                                                             
| screeningVoteMulti                          | 2532            | 156117 | 115032 | 477842 | 5       |                                                                                                             
| startNewDistributionPeriod                  | 642             | 48587  | 53289  | 76155  | 18      |                                                                                                                                                                                                                                                                                                                                                                                                                                                 
```


src/grants/GrantFund.sol
```diff
@@ -135,6 +135,41 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
         }
     }
 
+    // TODO: investigate passing proposalId instead of Proposal struct to save gas
```
EdNoepel: Seems like `proposalId` is passed in the `voteParams_` arg, and that this comment is stale.

src/grants/base/StandardFunding.sol
```diff
@@ -540,8 +546,10 @@ abstract contract StandardFunding is Funding, IStandardFunding {
         // record voters vote
         hasVotedScreening[proposal_.distributionId][account_] = true;
 
-        // vote for the given proposal
-        return super._castVote(proposal_.proposalId, account_, 1, "", "");
+        // emit VoteCast instead of VoteCastWithParams to maintain compatibility with Tally
+        emit VoteCast(account_, proposalId, 1, votes_, "");
```
EdNoepel: Why don't we need to call `super._castVote` anymore?

EdNoepel (commented):
A few comments/questions inline.

src/grants/base/StandardFunding.sol
MikeHathaway: Similar to `_fundingVote` our implementation changed such that we want to allow people to cast partial votes on a proposal without allocating their full voting power. `super._castVote` automatically allocates the full voting power. It also makes additional calls that aren't needed (like `_countVote` which is handled by us)

src/grants/GrantFund.sol
MikeHathaway: Had left it there as a pointer to check if there was any efficiencies gained from passing `proposalId` into `_screeningVote` and loading the `Proposal` struct from storage in there vs in the outer function but on further investigation the gains seem minimal and not worth the change. I removed the comment.

src/grants/GrantFund.sol
```diff
@@ -112,26 +112,33 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
...
-        if (voter.votingPower == 0) {
-            voter.votingPower          = Maths.wpow(_getVotesSinceSnapshot(msg.sender, screeningStageEndBlock - VOTING_POWER_SNAPSHOT_DELAY, screeningStageEndBlock), 2);
-            voter.remainingVotingPower = voter.votingPower;
-        }
+        // check that the funding stage is active
+        if (block.number > screeningStageEndBlock && block.number <= currentDistribution.endBlock) {
+
+            // this is the first time a voter has attempted to vote this period
+            if (voter.votingPower == 0) {
+                voter.votingPower          = Maths.wpow(_getVotesSinceSnapshot(msg.sender, screeningStageEndBlock - VOTING_POWER_SNAPSHOT_DELAY, screeningStageEndBlock), 2);
```
grandizzy: small gas improvement by avoiding one read from storage: declare
```
uint256 newVotingPower = Maths.wpow(_getVotesSinceSnapshot(msg.sender, screeningStageEndBlock - VOTING_POWER_SNAPSHOT_DELAY, screeningStageEndBlock), 2);
voter.votingPower  = newVotingPower;
voter.remainingVotingPower = newVotingPower;
```

same can be applied at https://github.com/ajna-finance/ecosystem-coordination/blob/70aa5c6169d897129d1de0833c5219f9d8850911/src/grants/GrantFund.sol#L215

src/grants/GrantFund.sol
```diff
@@ -186,10 +193,11 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
             QuarterlyDistribution storage currentDistribution = distributions[proposal.distributionId];
             uint256 screeningStageEndBlock = _getScreeningStageEndBlock(currentDistribution);
 
+            // check that the proposal is part of the current distribution period
```
grandizzy, line 212: this check can be moved after line 192 as it will avoid reading `currentDistribution` and calculating `screeningStageEndBlock` if reverts with InvalidVote, also `proposal.distributionId` can be loaded only once as

```
            Proposal storage proposal = standardFundingProposals[proposalId_];
            uint256 distributionId = proposal.distributionId;

            // check that the proposal is part of the current distribution period
            if (distributionId != distributionIdCheckpoints.latest()) revert InvalidVote();

            QuarterlyDistribution storage currentDistribution = distributions[distributionId];
            uint256 screeningStageEndBlock = _getScreeningStageEndBlock(currentDistribution);
```

src/grants/GrantFund.sol
```diff
@@ -112,26 +112,33 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
         QuadraticVoter storage voter = quadraticVoters[currentDistribution.id][msg.sender];
         uint256 screeningStageEndBlock = _getScreeningStageEndBlock(currentDistribution);
 
-        // this is the first time a voter has attempted to vote this period
-        if (voter.votingPower == 0) {
-            voter.votingPower          = Maths.wpow(_getVotesSinceSnapshot(msg.sender, screeningStageEndBlock - VOTING_POWER_SNAPSHOT_DELAY, screeningStageEndBlock), 2);
-            voter.remainingVotingPower = voter.votingPower;
-        }
+        // check that the funding stage is active
+        if (block.number > screeningStageEndBlock && block.number <= currentDistribution.endBlock) {
```
grandizzy: do we want to inform user that something wrong here like
`if (block.number <= screeningStageEndBlock || block.number > currentDistribution.endBlock) revert FundingStageInactive()`
or just silently do nothing if funding stage inactive?

grandizzy (commented):
overall LGTM, I am not fully knowledgeable on the voting mechanism so not sure if you want a different opinion / review on it. Have couple of suggestions for optimizing code, pls see inline

src/grants/GrantFund.sol
MikeHathaway: Updated at: https://github.com/ajna-finance/ecosystem-coordination/pull/44/commits/e3a4d17cec7deac6d8dada0657f2f2a2fb5beb5b

src/grants/GrantFund.sol
MikeHathaway: updated at: https://github.com/ajna-finance/ecosystem-coordination/pull/44/commits/e3a4d17cec7deac6d8dada0657f2f2a2fb5beb5b

src/grants/GrantFund.sol
MikeHathaway: Yeah that might make sense - i'd been avoiding that to minimize the difference between the voteMulti and `_castVote` functions. Can't really do that revert in `_castVote` as it's needed to override the various voting methods in `Governor` and is the means by which the different voting mechanisms are selected

prat-gpt: I am in favour of adding revert here as in the `_castVote` the voting would be done, either in screening or in funding. Here if we do not revert the transaction will pass but will not vote anywhere, revert would be a better option. 
I think adding revert will minimize the difference between voteMulti and _castVote more when compared with not adding a revert because then in both the cases if a transaction is successful that would mean voting was done.

src/grants/GrantFund.sol
```diff
@@ -149,24 +192,34 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
         // standard funding mechanism
```
prat-gpt, line 206: can update the params_ description comment on line 186 from `The amount of votes being allocated in the funding stage.` to `The amount of votes being allocated in the screening/funding stage`.

src/grants/base/StandardFunding.sol
```diff
@@ -500,28 +508,30 @@ abstract contract StandardFunding is Funding, IStandardFunding {
...
+            currentTopTenProposals.push(proposalId);
 
             // sort top ten proposals
             _insertionSortProposalsByVotes(currentTopTenProposals);
         }
         else {
             // proposal is already in the array
             if (indexInArray != -1) {
-                currentTopTenProposals[uint256(indexInArray)] = proposal_.proposalId;
+                currentTopTenProposals[uint256(indexInArray)] = proposalId;
```
prat-gpt: this seems like a redundant line of code to me. Can we remove this?

src/grants/GrantFund.sol
```diff
@@ -276,7 +329,7 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
 
             // screening stage
             if (block.number >= currentDistribution.startBlock && block.number <= screeningStageEndBlock) {
-                hasVoted_ = hasVotedScreening[proposal.distributionId][account_];
+                hasVoted_ = screeningVotesCast[proposal.distributionId][account_] != 0;
```
prat-gpt, line 387: should we update `hasVoted_` to true when `screeningVotesCast` by a user are equal to his full voting power and he cannot vote anymore. Maybe this change will make it more compatible with tally because as of now if someone votes partially tally will show that he has finished voting and can no longer vote.

prat-gpt (approved):
LGTM. Just added some comments.

src/grants/base/StandardFunding.sol
prat-gpt: I noticed that this has been resolved in George's improvements PR.

src/grants/GrantFund.sol
MikeHathaway: Will update

src/grants/GrantFund.sol
MikeHathaway: That's a good question. It would be nice to maintain compatibility but the tally implementation really expects max power on a given proposal which is mechanistically different from us. This would also raise the question of whether we should return false if they haven't used their full voting power in the funding round which is *very* different from tallys expectations of the mechanism

src/grants/GrantFund.sol
MikeHathaway: The problem is `_castVote` needs to support every single voting mechanism. It isn't possible to revert in one of the cases as it could be meant to be a different stage or mechanism. We would have to do a wholesale rewrite to add it to `_castVote` which i'm not sure is possible while still overriding Governor. We could add it to the `multi` methods but they would then be inconsistent with singular votes done through `_castVote`

grandizzy (approved):

src/grants/base/ExtraordinaryFunding.sol
```diff
@@ -37,30 +37,37 @@ abstract contract ExtraordinaryFunding is Funding, IExtraordinaryFunding {
...
         proposalId_ = hashProposal(targets_, values_, calldatas_, descriptionHash_);
 
         ExtraordinaryFundingProposal storage proposal = extraordinaryFundingProposals[proposalId_];
 
-        if (proposal.executed) {
-            revert ExecuteExtraordinaryProposalInvalid();
-        }
+        uint256 tokensRequested = uint256(proposal.tokensRequested);
+
+        // revert if executed or if the proposal has received more votes than minimumThreshold and tokensRequestedPercentage of all tokens
```
dmitriia, line 52: Comment to be:
```
// revert if executed or if the proposal has received less votes than minimumThreshold of non-treasury tokens plus tokensRequested
```

dmitriia (commented):
Ideally all explicit casting, here it is uint128 and uint112, to be replaced with safe casting.

PR #47, opened 2023-03-02
Improvements

grandizzy:
- Apply same code style
- speed up tests by forking from specific block
- gas improvements by reducing number of storage loads
- remove unused return of votes from _screeningVote function
-  use internal _getSliceOfNonTreasury and _getSliceOfTreasury functions, change public to external for getSliceOfNonTreasury and getSliceOfTreasury functions
- packed structs

WITH CHANGES:

```                                                                                           
| castVote                                    | 1402            | 33546  | 33738  | 35738  | 283     |                                                                                                             
| castVoteWithReasonAndParams                 | 1911            | 75935  | 74404  | 125302 | 491     |                                                                                                             
| checkSlate                                  | 1028            | 45028  | 61002  | 78933  | 16      |                                                                                                             
| claimDelegateReward                         | 1482            | 57876  | 58836  | 65636  | 229     |                                                                                                             
| execute                                     | 2457            | 2457   | 2457   | 2457   | 1       |                                                                                                             
| executeExtraordinary                        | 7343            | 65608  | 65511  | 121258 | 9       |                                                                                                             
| executeStandard                             | 9556            | 46292  | 63232  | 72987  | 8       |                                                                                                             
| fundingVotesMulti                           | 105054          | 182141 | 153813 | 396896 | 7       |                                                                                                                                                                   
| propose                                     | 2944            | 2944   | 2944   | 2944   | 1       |                                                                                                             
| proposeExtraordinary                        | 4739            | 63345  | 82007  | 88346  | 14      |                                                                                                             
| proposeStandard                             | 4695            | 79480  | 82411  | 84750  | 93      |                                                                                                             
| screeningVoteMulti                          | 2532            | 156117 | 115032 | 477842 | 5       |                                                                                                             
| startNewDistributionPeriod                  | 1011            | 51897  | 57658  | 77124  | 18      |                                                                                                             
```

BEFORE CHANGES:
```
| castVote                                    | 1406            | 34224  | 33609  | 59509  | 181     |
| castVoteWithReasonAndParams                 | 2185            | 81941  | 102577 | 167891 | 827     |
| checkSlate                                  | 906             | 45854  | 61706  | 79619  | 16      |
| claimDelegateReward                         | 1613            | 58362  | 58911  | 65711  | 397     |
| execute                                     | 2349            | 2349   | 2349   | 2349   | 1       |
| executeExtraordinary                        | 7125            | 81511  | 85437  | 141190 | 9       |
| executeStandard                             | 10841           | 60675  | 85263  | 95017  | 8       |
| fundingVotesMulti                           | 127669          | 217711 | 174084 | 497450 | 7       |
| proposeExtraordinary                        | 4468            | 78779  | 103720 | 110059 | 14      |
| proposeStandard                             | 4510            | 78694  | 83198  | 85198  | 75      |
| screeningVoteMulti                          | 3210            | 229206 | 162491 | 744884 | 5       |
| startNewDistributionPeriod                  | 1618            | 140313 | 152376 | 167334 | 18      |
```

### DIFF
--- a/src/grants/base/ExtraordinaryFunding.sol
+++ b/src/grants/base/ExtraordinaryFunding.sol
@@ -37,7 +37,12 @@ abstract contract ExtraordinaryFunding is Funding, IExtraordinaryFunding {
     /**************************/
 
     /// @inheritdoc IExtraordinaryFunding
-    function executeExtraordinary(address[] memory targets_, uint256[] memory values_, bytes[] memory calldatas_, bytes32 descriptionHash_) external nonReentrant returns (uint256 proposalId_) {
+    function executeExtraordinary(
+        address[] memory targets_,
+        uint256[] memory values_,
+        bytes[] memory calldatas_,
+        bytes32 descriptionHash_
+    ) external nonReentrant returns (uint256 proposalId_) {
         proposalId_ = hashProposal(targets_, values_, calldatas_, descriptionHash_);
 
         ExtraordinaryFundingProposal storage proposal = extraordinaryFundingProposals[proposalId_];
@@ -115,7 +120,10 @@ abstract contract ExtraordinaryFunding is Funding, IExtraordinaryFunding {
      * @param  account_    The voting account.
      * @return votes_      The amount of votes cast.
      */
-    function _extraordinaryFundingVote(uint256 proposalId_, address account_) internal returns (uint256 votes_) {
+    function _extraordinaryFundingVote(
+        uint256 proposalId_,
+        address account_
+    ) internal returns (uint256 votes_) {
         if (hasVotedExtraordinary[proposalId_][account_]) revert AlreadyVoted();
 
         ExtraordinaryFundingProposal storage proposal = extraordinaryFundingProposals[proposalId_];
@@ -139,7 +147,9 @@ abstract contract ExtraordinaryFunding is Funding, IExtraordinaryFunding {
      * @param  proposalId_ The ID of the proposal being checked.
      * @return             Boolean indicating whether the proposal has succeeded.
      */
-    function _extraordinaryFundingVoteSucceeded(uint256 proposalId_) internal view returns (bool) {
+    function _extraordinaryFundingVoteSucceeded(
+        uint256 proposalId_
+    ) internal view returns (bool) {
         return extraordinaryFundingProposals[proposalId_].succeeded;
     }
 
@@ -168,7 +178,9 @@ abstract contract ExtraordinaryFunding is Funding, IExtraordinaryFunding {
      * @param percentage_ The percentage of the Non treasury to retrieve, in WAD.
      * @return The number of tokens, in WAD.
      */
-    function getSliceOfNonTreasury(uint256 percentage_) public view returns (uint256) {
+    function getSliceOfNonTreasury(
+        uint256 percentage_
+    ) public view returns (uint256) {
         uint256 totalAjnaSupply = IERC20(ajnaTokenAddress).totalSupply();
         return Maths.wmul(totalAjnaSupply - treasury, percentage_);
     }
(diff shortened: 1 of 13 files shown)


src/grants/base/ExtraordinaryFunding.sol
```diff
@@ -178,21 +194,47 @@ abstract contract ExtraordinaryFunding is Funding, IExtraordinaryFunding {
...
+    }
+
     /// @inheritdoc IExtraordinaryFunding
-    function getExtraordinaryProposalInfo(uint256 proposalId_) external view returns (uint256, uint256, uint256, uint256, uint256, bool, bool) {
-        ExtraordinaryFundingProposal memory proposal = extraordinaryFundingProposals[proposalId_];
+    function getExtraordinaryProposalInfo(
+        uint256 proposalId_
+    ) external view override returns (uint256, uint256, uint256, uint256, uint256, bool, bool) {
         return (
-            proposal.proposalId,
```
MikeHathaway, line 189: wouldn't it be cheaper for an external contract caller if it was left as is? This would be multiple `SLOAD` vs multiple `MLOAD` before

grandizzy: hm, not sure if there's any difference for an external view call?

MikeHathaway: My understanding is that it wouldn't make a difference if it was called by RPC to a local node, but if an external contract was created that called this function as part of a state changing operation (like a VoteManager contract) than it would be higher (unless view functions trigger automatic loads into memory as opposed to storage)

grandizzy: afaik view functions guarantee storage is not altered so they load everything in memory. needs to be double checked :)

MikeHathaway (approved):
LGTM

src/grants/base/ExtraordinaryFunding.sol
MikeHathaway: Gotcha, yeah looks good then

MikeHathaway:
Added a bunch of extra struct packing, which further reduced gas. For reference on int sizes: https://velvetshark.com/articles/max-int-values-in-solidity. Screening stage + extraordinary WAD numbers require at least uint96, while funding stage WAD numbers require at least uint120

src/grants/interfaces/IExtraordinaryFunding.sol
```diff
@@ -36,10 +31,10 @@ interface IExtraordinaryFunding {
      */
     struct ExtraordinaryFundingProposal {
         uint256  proposalId;      // Unique ID of the proposal. Hashed from proposeExtraordinary inputs.
-        uint256  tokensRequested; // Number of AJNA tokens requested.
-        uint256  startBlock;      // Block number of the start of the extraordinary funding proposal voting period.
-        uint256  endBlock;        // Block number of the end of the extraordinary funding proposal voting period.
-        uint256  votesReceived;   // Total votes received for this proposal.
+        uint128  startBlock;      // Block number of the start of the extraordinary funding proposal voting period.
+        uint128  endBlock;        // Block number of the end of the extraordinary funding proposal voting period.
+        uint128  tokensRequested; // Number of AJNA tokens requested.
```
prat-gpt, line 36: maybe we can make `votesReceived` as uint128 and `tokensRequested` as unit112 as I think `votesReveived` will be most of the time greater than `tokensRequested`

grandizzy: makes sense also considering that Ajna total supply is 1b so `tokensRequested` cannot exceed that

prat-gpt (approved):
LGTM!. Added one comment.

src/grants/interfaces/IExtraordinaryFunding.sol
prat-gpt: Yes right, 1B = 10**9, we will need to store at max 10**27, which can be easily stored in both uint112 or uint128 so it will work without any issue if we keep the struct to current state as well.

PR #49, opened 2023-03-04
remove transferFromWithPermit

MikeHathaway:
Remove `transferFromWithPermit` method. As implemented, it was vulnerable to front running attacks. `permit` calls should be implemented at the spender itself, so the spender would atomically call `permit` as part of a larger operation. This prevents someone from stripping the signature our of the mempool and updating the `to` address.

Since no custom logic remains around permit, the test was removed entirely. `permit` is still available for implementation by others via `ERC20Permit`, which is tested by OZ.


grandizzy:
per discussions with ToB, this can be closed, is that correct @MikeHathaway ?

MikeHathaway:
Closed in favor of the deployed version

PR #51, opened 2023-03-09
Invariants Testing

MikeHathaway:
Add tests for the majority of the standard funding mechanism's invariants. 
- Created scenarios for each portion of the distribution period end to end, with tests for each period.
- Update test/ directory structure

There are a lot of outstanding todos which I'm looking to fix over the next few days, ideally in a separate PR. Currently being worked:
- Global GrantFund invariants
- More sophisticated time manipulation

A list of the completion status of the Standard invariants is available here: https://www.notion.so/worlds-best-crypto-lending-platform/GrantFund-Standard-Funding-Mechanism-invariant-testing-8b87a0a814a5405aa42f931590d7a332?pvs=4


foundry.toml
```diff
@@ -12,4 +12,13 @@ fork_block_number   = 16_295_000
...
 [fuzz]
-runs = 150
\ No newline at end of file
+runs = 150
+
+[invariant]
+runs           = 1     # The number of calls to make in the invariant tests
+depth          = 500   # The number of times to run the invariant tests
+call_override  = false # Override calls
+fail_on_revert = true  # Fail the test if the contract reverts
```
ith-harvey, line 21: I'm personally a fan of fail on revert = true as that was the original instruction provided by Lucas.

However, in the event where we are not making enough progress due to our timeline it may make sense to set this to false and run using the number of runs vs reverts as a goalpost just like we are doing in contracts repo

test/invariants/INVARIANTS.md
```diff
@@ -0,0 +1,70 @@
+# GrantFund Invariants
+
+## Grant Fund Invariants:
+   - **GF1**: Unused proposal states should always be 0.
+   - **GF2**: Treasury should always be less than or equal to the contract's token blance.
```
ith-harvey: Unclear on what this invariant is saying... is Treasury in this context a variable? If so we should use `treasury` so it's a bit clearer?

test/invariants/base/StandardTestBase.sol
```diff
@@ -0,0 +1,40 @@
...
+            NUM_ACTORS,
+            tokensNotInTreasury
+        );
+
+        // explicitly target handler
+        targetContract(address(_standardHandler));
+
+        // skip time for snapshots and start distribution period
+        vm.roll(block.number + 100);
+        // vm.rollFork(block.number + 100);
```
ith-harvey: is this comment needed? suggest removing

test/invariants/base/TestBase.sol
```diff
@@ -0,0 +1,52 @@
...
+    uint256 treasury = 500_000_000 * 1e18;
+
+    function setUp() public virtual {
+        vm.createSelectFork(vm.envString("ETH_RPC_URL"), _startBlock);
+
+        // provide cheatcode access to the standard funding handler
+        vm.allowCheatcodes(0x4447E7a83995B5cCDCc9A6cd8Bc470305C940DA3);
+
+        // Ajna Token contract address on mainnet
+        _token = IAjnaToken(0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079);
```
ith-harvey: suggest renaming to `_ajna` to match contract repo

test/invariants/handlers/Handler.sol
```diff
@@ -0,0 +1,139 @@
...
+import { IAjnaToken }          from "../../utils/IAjnaToken.sol";
+import { GrantFundTestHelper } from "../../utils/GrantFundTestHelper.sol";
+
+import { GrantFund } from "../../../src/grants/GrantFund.sol";
+import { IStandardFunding } from "../../../src/grants/interfaces/IStandardFunding.sol";
+
+contract Handler is Test, GrantFundTestHelper {
+
+    // state variables
+    IAjnaToken        internal  _token;
```
ith-harvey: suggest renaming `_token` -> `_ajna`

test/invariants/handlers/Handler.sol
```diff
@@ -0,0 +1,139 @@
...
+    address[] public actors;
+    address _tokenDeployer;
+
+    // logging
+    mapping(bytes32 => uint256) public numberOfCalls;
+
+    // randomness counter
+    uint256 internal counter = 1;
+
+    constructor(address payable grantFund_, address token_, address tokenDeployer_, uint256 numOfActors_, uint256 tokensToDistribute_) {
```
ith-harvey: suggest new line after every arg to clean up formatting

test/invariants/handlers/Handler.sol
```diff
@@ -0,0 +1,139 @@
...
+        _votingToken = IVotes(address(_token));
+
+        // deploy growth fund contract
+        _grantFund = GrantFund(grantFund_);
+
+        // token deployer
+        _tokenDeployer = tokenDeployer_;
+
+        // instantiate actors
+        actors = _buildActors(numOfActors_, tokensToDistribute_);
```
ith-harvey, line 65: Due to the size the contracts repo, the one off excpetions in certain scenarios and the short timelines we decided to go against supplying tokens to actors on every call and went with supplying them the tokens at in the setup.

I don't much like that approach and would suggest we provide tokens to the actor performing the call when the action is called. I think this results in less errors and more depth of calls. More difficult at first but easier in the later stages of dev

Please see how maple does it -> https://github.com/maple-labs/maple-core-v2/blob/main/tests/invariants/actors/LpHandler.sol#L79

test/invariants/handlers/StandardHandler.sol
```diff
@@ -0,0 +1,804 @@
...
+
+    /*********************/
+    /*** SFM Functions ***/
+    /*********************/
+
+    function startNewDistributionPeriod(uint256 actorIndex_) external useRandomActor(actorIndex_) returns (uint24 newDistributionId_) {
+        numberOfCalls['SFH.startNewDistributionPeriod']++;
+        systemTime++;
+
+        // vm.roll(block.number + 100);
```
ith-harvey: is this comment still needed?

test/invariants/handlers/StandardHandler.sol
```diff
@@ -0,0 +1,804 @@
...
+        // vm.rollFork(block.number + 100);
+
+        try _grantFund.startNewDistributionPeriod() returns (uint24 newDistributionId) {
+            newDistributionId_ = newDistributionId;
+            // FIXME: remove this
```
ith-harvey: not sure why this exists / what issue this FixMe handles

test/invariants/handlers/StandardHandler.sol
```diff
@@ -0,0 +1,804 @@
...
+
+        // generate proposal params
+        (
+            address[] memory targets,
+            uint256[] memory values,
+            bytes[] memory calldatas,
+            string memory description
+        ) = generateProposalParams(testProposalParams);
+
+        console.log("description", description);
```
ith-harvey: log still needed? recommend removing

test/invariants/StandardFinalizeInvariant.t.sol
```diff
@@ -0,0 +1,147 @@
...
+        selectors[2] = _standardHandler.executeStandard.selector;
+
+        // ensure utility functions are excluded from the invariant runs
+        targetSelector(FuzzSelector({
+            addr: address(_standardHandler),
+            selectors: selectors
+        }));
+    }
+
+    function invariant_CS1_CS2_CS3_CS4_CS5() external {
```
ith-harvey: I like the idea of breaking each invariant into their own separate test even if the function remains a one liner... much like how Maple does it and I suggest we do it here too. Think it results in cleaner code with less comments, easier to follow IMO -> https://github.com/maple-labs/maple-core-v2/blob/00f01ae7175885f8d49ac201a1c72465e320b2f6/tests/invariants/DefaultsInvariants.t.sol#L66

ith-harvey (commented):
Would like to see the invariants in README rewritten using the actual variable names in the contracts to more closely match the `contracts` repo invariants as well as provide more clarity to auditors / code reviewers. Happy to help with that.

Alot of comments throughout the code... seeing as there is some future work and this is a WIP I think FIXME's are fine but random one off comments without any context are confusing, I suggest removing those if and where possible. Didn't flag all of them as there were too many.

Structure looks decent, once some of the structural inquiries are addressed (either a response or change) happy to approve so we can get this scaffolding in.

test/invariants/handlers/Handler.sol
MikeHathaway: That isn't possible here as actors need the tokens delegated by certain blocks to be able to take actions. Vote counts would be 0

test/invariants/StandardFinalizeInvariant.t.sol
```diff
@@ -0,0 +1,181 @@
...
+
+        // check proposal state of the constituents of the top slate
+        uint256 totalTokensRequested = 0;
+        for (uint256 i = 0; i < topSlateProposalIds.length; ++i) {
+            uint256 proposalId = topSlateProposalIds[i];
+            (, , , uint128 tokensRequested, int128 fundingVotesReceived, ) = _grantFund.getProposalInfo(proposalId);
+            totalTokensRequested += tokensRequested;
+
+            // invariant CS3: Proposal slate should never contain a proposal with negative funding votes received
+            assertTrue(fundingVotesReceived >= 0);
```
grandizzy: for invariants we may want to use `require` with specific revert reason, like `require(fundingVotesReceived >= 0, "CS3: proposal slate contains proposal with negative funding")` or something along these lines

test/invariants/INVARIANTS.md
```diff
@@ -0,0 +1,70 @@
+# GrantFund Invariants
+
+## Grant Fund Invariants:
+   - **GF1**: Unused proposal states should always be 0.
```
grandizzy, line 4: I don't find where these 2 invariants are checked in code, can you clarify?

test/invariants/INVARIANTS.md
```diff
@@ -0,0 +1,70 @@
...
+   - **GF2**: Treasury should always be less than or equal to the contract's token blance.
+<!-- 
+   - TODO: add invariants around treasury balance post updates and with partial slate executions
+ -->
+
+## Standard Funding Mechanism Invariants
+
+- #### Distribution Period:
```
grandizzy, line 12: same as above, I cannot find in code where these invariants are checked

test/invariants/StandardScreeningInvariant.t.sol
```diff
@@ -0,0 +1,106 @@
...
+            addr: address(_standardHandler),
+            selectors: selectors
+        }));
+
+    }
+
+    function invariant_SS1_SS3_SS4_SS5_SS7() public {
+        uint256[] memory topTenProposals = _grantFund.getTopTenProposals(_grantFund.getDistributionId());
+
+        // invariant: 10 or less proposals should make it through the screening stage
```
grandizzy: would be nice to mark it as `SS1`

test/invariants/StandardScreeningInvariant.t.sol
```diff
@@ -0,0 +1,106 @@
...
+        assertTrue(topTenProposals.length <= 10);
+
+        if (_standardHandler.screeningVotesCast() > 0) {
```
grandizzy: if I am not wrong this one is `SS2`, we should mark it as such

test/invariants/StandardScreeningInvariant.t.sol
```diff
@@ -0,0 +1,106 @@
...
+        uint256 standardFundingProposalsSubmitted = _standardHandler.standardFundingProposalCount();
+
+        // check invariants against all submitted proposals
+        for (uint256 j = 0; j < standardFundingProposalsSubmitted; ++j) {
+            (, , uint256 votesReceived, , , ) = _grantFund.getProposalInfo(_standardHandler.standardFundingProposals(j));
+            // invariant SS4: votes recieved for a proposal can only be positive
+            assertTrue(votesReceived >= 0);
+
+            // invariant SS7: a proposal should never receive more votes than the token supply.
+            assertTrue(votesReceived <= 1_000_000_000 * 1e18);
```
grandizzy: token supply can be less than 1bn in case reserve auction taken. I suggest to replace this with `token.totalSupply()` call and maybe add a fuzz call for burning ajna tokens

test/invariants/StandardFundingInvariant.t.sol
```diff
@@ -0,0 +1,117 @@
...
+        targetSelector(FuzzSelector({
+            addr: address(_standardHandler),
+            selectors: selectors
+        }));
+    }
+
+    function invariant_FS1_FS2_FS3() external {
+        uint256[] memory topTenProposals = _grantFund.getTopTenProposals(_grantFund.getDistributionId());
+
+        // invariant: 10 or less proposals should make it through the screening stage
```
grandizzy: looks like `SS1`?

test/invariants/StandardFundingInvariant.t.sol
```diff
@@ -0,0 +1,117 @@
...
+        assertTrue(topTenProposals.length <= 10);
```
grandizzy, line 48: can use `assertLe` and `assertGe` like shown in https://book.getfoundry.sh/forge/invariant-testing#defining-invariants or the `require` with invariant specified in revert reason (I think 2nd approach is better / cleaner)

test/invariants/StandardFundingInvariant.t.sol
```diff
@@ -0,0 +1,117 @@
...
+
+            if (votingPower != 0 && remainingVotingPower == 0) {
+                assertTrue(numberOfProposalsVotedOn == fundingVotesCast.length);
+                assertTrue(numberOfProposalsVotedOn > 0);
+            }
+
+            // invariant FS8: a voter should never be able to cast more votes than the Ajna token supply of 1 billion.
+            assertTrue(uint256(_standardHandler.sumFundingVotes(fundingVoteParams)) <= 1_000_000_000 * 1e18);
+
+            // TODO: check getFundingPowerVotes to see if remaining voting power matches expectations
```
grandizzy: why the TODO and next line commented?

test/invariants/StandardScreeningInvariant.t.sol
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        uint256 standardFundingProposalsSubmitted = _standardHandler.standardFundingProposalCount();
+
+        // check invariants against all submitted proposals
+        for (uint256 j = 0; j < standardFundingProposalsSubmitted; ++j) {
+            (, , uint256 votesReceived, , , ) = _grantFund.getProposalInfo(_standardHandler.standardFundingProposals(j));
+            // invariant SS4: votes recieved for a proposal can only be positive
+            assertTrue(votesReceived >= 0);
```
grandizzy: should invariant be `votesReceived >= 0` or strictly greater than 0 `votesReceived > 0` ?

test/invariants/base/TestBase.sol
grandizzy: for `contracts` repo we introduced a `burnableToken` and use that instead forking mainnet and Ajna. This makes tests faster and avoid being rate limited by infura, see https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC20Pool/invariants/base/BaseHandler.sol#L84

test/invariants/handlers/Handler.sol
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    // modifier useBlock() {
```
grandizzy: can this modifier and `setCurrentBlock` function below be removed?

test/invariants/handlers/StandardHandler.sol
```diff
@@ -0,0 +1,857 @@
...
+        catch (bytes memory _err){
+            bytes32 err = keccak256(_err);
+            require(
+                err == keccak256(abi.encodeWithSignature("ExecuteProposalInvalid()")) ||
+                err == keccak256(abi.encodeWithSignature("ProposalNotSuccessful()"))
+            );
+        }
+    }
+
+    function claimDelegateReward(uint256 actorIndex_, uint256 proposalToExecute_) external useRandomActor(actorIndex_) {
```
grandizzy: `proposalToExecute_` is not used

test/invariants/handlers/StandardHandler.sol
```diff
@@ -0,0 +1,857 @@
...
+            numberOfCalls["SFH.executeStandard"]
+        );
+    }
+
+    function logProposalSummary() external view {
+        console.log("\nProposal Summary\n");
+        console.log("Number of Proposals", standardFundingProposalCount);
+        for (uint256 i = 0; i < standardFundingProposalCount; ++i) {
+            console.log("------------------");
+            (uint256 proposalId, uint24 distributionId, uint128 votesReceived, uint128 tokensRequested, int128 fundingVotesReceived, bool executed) = _grantFund.getProposalInfo(standardFundingProposals[i]);
```
grandizzy: we may want to log the `fundingVotesReceived` too (and get rid of compiler warning)?

test/invariants/handlers/StandardHandler.sol
```diff
@@ -0,0 +1,857 @@
...
+        }
+    }
+
+    function sumFundingVotes(IStandardFunding.FundingVoteParams[] memory fundingVotes_) public pure returns (int256 sum_) {
+        for (uint256 i = 0; i < fundingVotes_.length; ++i) {
+            sum_ += Maths.abs(fundingVotes_[i].votesUsed);
+        }
+    }
+
+    function countNegativeFundingVotes(address actor_, IStandardFunding.FundingVoteParams[] memory fundingVotes_) public pure returns (uint256 count_) {
```
grandizzy: `actor_` is not used, I think we'd want to check votes used per actor here?

test/invariants/StandardFinalizeInvariant.t.sol
```diff
@@ -0,0 +1,181 @@
...
+        // invariant CS1: total tokens requested should be <= 90% of fundsAvailable
+        assertTrue(totalTokensRequested <= uint256(fundsAvailable) * 9 / 10);
+
+        // invariant CS5: proposal slate should never contain duplicate proposals
+        assertFalse(_standardHandler.hasDuplicates(topSlateProposalIds));
+    }
+
+    function invariant_ES1_ES3() external {
+        uint24 distributionId = _grantFund.getDistributionId();
+        (, , , uint128 fundsAvailable, , bytes32 topSlateHash) = _grantFund.getDistributionPeriodInfo(distributionId);
```
grandizzy: `fundsAvailable` is not used, should we remove or assert invariant `CS1` here?

test/invariants/StandardFinalizeInvariant.t.sol
```diff
@@ -0,0 +1,181 @@
...
+                assertTrue(delegationRewardsClaimed >= 0);
+
+                // invariant DR2: Delegation rewards are 0 if voter didn't vote in both stages.
+                assertTrue(fundingVoteParams.length > 0 && screeningVoteParams.length > 0);
+            }
+        }
+
+        // invariant DR1: Cumulative delegation rewards should be 10% of a distribution periods GBC.
+        assertTrue(totalRewardsClaimed <= fundsAvailable * 1 / 10);
+        // assertTrue(totalRewardsClaimed <= fundsAvailable * 9 / 10);
```
grandizzy: which assert is the right one, the one above or this one that is commented out?

test/invariants/StandardFinalizeInvariant.t.sol
```diff
@@ -0,0 +1,181 @@
...
+        uint24 distributionId = _grantFund.getDistributionId();
+        (, , , uint128 fundsAvailable, , ) = _grantFund.getDistributionPeriodInfo(distributionId);
+
+        uint256 totalRewardsClaimed;
+
+        for (uint256 i = 0; i < _standardHandler.getActorsCount(); ++i) {
+            address actor = _standardHandler.actors(i);
+
+            // get the initial funding stage voting power of the actor
+            (uint128 votingPower, uint128 remainingVotingPower, uint256 numberOfProposalsVotedOn) = _grantFund.getVoterInfo(distributionId, actor);
```
grandizzy: `votingPower` is not used, I suspect we want to assert it? Same for `remainingVotingPower` and ``numberOfProposalsVotedOn``, should we check total voting power equals remaining voting power + numberOfProposalsVotedOn?

test/invariants/StandardFinalizeInvariant.t.sol
```diff
@@ -0,0 +1,181 @@
...
+    function invariant_call_summary() external view {
+        uint24 distributionId = _grantFund.getDistributionId();
+
+        _standardHandler.logCallSummary();
+        _standardHandler.logProposalSummary();
+        _logFinalizeSummary(distributionId);
+    }
+
+    function _logFinalizeSummary(uint24 distributionId_) internal view {
+        (, , , uint128 fundsAvailable, , bytes32 topSlateHash) = _grantFund.getDistributionPeriodInfo(distributionId_);
```
grandizzy, line 230: `fundsAvailable` is not used, do we want to log it?

test/invariants/handlers/StandardHandler.sol
```diff
@@ -0,0 +1,857 @@
...
+
+        try _grantFund.startNewDistributionPeriod() returns (uint24 newDistributionId) {
+            newDistributionId_ = newDistributionId;
+            // FIXME: remove this
+            vm.roll(block.number + 100);
+            vm.rollFork(block.number + 100);
+        }
+        catch (bytes memory _err){
+            bytes32 err = keccak256(_err);
+            require(
```
grandizzy, line 79: not sure if worth changing here but for `contracts` we have defined `_ensurePoolError` function with all known pool errors https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC20Pool/invariants/base/BaseHandler.sol#L185 At minimum would be nice to have the `"Unexpected revert error"` reason

grandizzy (changes requested):
looks good for the first iteration (noticed several TODOs), had couple of comments / suggestions.

test/invariants/StandardFinalizeInvariant.t.sol
MikeHathaway: Yeah I like that idea. I will update it accordingly

test/invariants/INVARIANTS.md
MikeHathaway: They aren't checked yet. In order to properly test distribution period I need to first improve the way we manipulate time. The system only functions properly when it's at certain block heights, so need to be able to progress between the stages, as well as between different distribution periods.

test/invariants/handlers/StandardHandler.sol
MikeHathaway: I agree about adding "Unexpected revert error". It's cleaner with `ensurePoolError`, but i'm concerned that some test granularity is lost without the checks for the specific reverts (i.e. only those should ever be seen, not a different pool error).

grandizzy: I think that granularity is already covered by unit tests but I'm OK with either ways

test/invariants/handlers/StandardHandler.sol
MikeHathaway: removed

test/invariants/handlers/StandardHandler.sol
MikeHathaway: removed

test/invariants/handlers/StandardHandler.sol
MikeHathaway: removed

test/invariants/handlers/Handler.sol
MikeHathaway: updated

test/invariants/base/StandardTestBase.sol
MikeHathaway: removed

test/invariants/StandardFinalizeInvariant.t.sol
MikeHathaway: added a log

test/invariants/handlers/StandardHandler.sol
MikeHathaway: removed

test/invariants/handlers/StandardHandler.sol
MikeHathaway: logged

test/invariants/handlers/StandardHandler.sol
MikeHathaway: removed

test/invariants/StandardFinalizeInvariant.t.sol
MikeHathaway: Updated to check the totals across all claimers

test/invariants/StandardFinalizeInvariant.t.sol
MikeHathaway: I need those values to assert invariant DR3. I'm still trying to figure out the best way to check that invariant

test/invariants/StandardFinalizeInvariant.t.sol
MikeHathaway: removed

test/invariants/handlers/Handler.sol
MikeHathaway: Updated, with an additional `roll` method. I'm still experimenting with the best approach for getting the system into the correct stage to enable the invariants to actually be checked, while also progressing over time

test/invariants/StandardScreeningInvariant.t.sol
MikeHathaway: Updated with `totalSupply`. Treating a fuzz test that burns ajna tokens as out of scope

test/invariants/StandardFundingInvariant.t.sol
MikeHathaway: Updated it as `FS1`. It's essentially the same invariant check as the subsequent assertion, just performed differently

test/invariants/StandardScreeningInvariant.t.sol
MikeHathaway: We currently allow zero votes. We can potentially start to restrict that.

test/invariants/StandardScreeningInvariant.t.sol
MikeHathaway: Updated

test/invariants/StandardScreeningInvariant.t.sol
MikeHathaway: Was a different check on `SS6` duplicated the comment above it

test/invariants/StandardFundingInvariant.t.sol
MikeHathaway: Updated to use `require` statements instead.

test/invariants/StandardFundingInvariant.t.sol
MikeHathaway: removed

test/invariants/INVARIANTS.md
MikeHathaway: These invariants aren't checked yet. The list of currently completed invariants is available here: https://www.notion.so/worlds-best-crypto-lending-platform/GrantFund-Standard-Funding-Mechanism-invariant-testing-8b87a0a814a5405aa42f931590d7a332?pvs=4

grandizzy (approved):
I'm OK with merging in this state as a first iteration, then improve, couple of things noticed
- Log.sol is entirely commented out, not sure if we want to keep it or not (for now I just changed it to prevent compiler warnings)
- `make tests` was executing invariants tests as well, added the `--mt test` to prevent such and minimize GH action time

test/invariants/INVARIANTS.md
MikeHathaway: Updated accordingly

test/invariants/base/TestBase.sol
MikeHathaway: Updated

test/invariants/handlers/Handler.sol
MikeHathaway: updated

test/invariants/StandardFinalizeInvariant.t.sol
MikeHathaway: The issue is we need a really high call depth for these tests to work properly, and i need to iterate across both every actor, and every proposal for a lot of these invariant checks. Splitting this into two iterations per invariant would make the test time prohibitive

foundry.toml
MikeHathaway: I've found that not having `revert=true` slowed me down as problems went unexposed

test/invariants/handlers/Handler.sol
ith-harvey: I see -> `Voting power in the screening stage is based upon a snapshot of an address' voting power 33 blocks prior to the screening stage’s start block, `

Agreed, resolving.

ith-harvey (approved):
LGTM

PR #54, opened 2023-03-22
Proposed changes:

grandizzy:
- code style: add sections, use _ for all internal state vars
- fundTreasury function to not return treasury value, emit event at the end after transfer done
- gas improvements read currentDistributionId only once in startNewDistributionPeriod and fundingVote functions


MikeHathaway (approved):
LGTM. Thanks for the changes

Issue #55, opened 2023-03-28
Attacker can use Extraordinary proposal to steal extraordinary amounts of Ajna


rmi7:
The `ExtraordinaryFunding` contract's `voteExtraordinary` function allows any address to be passed in as the `account` that will place a vote. This allows an attacker to create a proposal and by calling `voteExtraordinary` for each account that has voting power, make any proposal succeed (as long as the minimum treshold is adhered to). For example, to transfer millions of Ajna tokens to the attacker's address.

```solidity
    function voteExtraordinary(
        address account_,
        uint256 proposalId_
    ) external override returns (uint256 votesCast_) {
        votesCast_ = _extraordinaryFundingVote(account_, proposalId_);
    }
```

To be able to place votes on both Standard and Extraordinary proposals, an account must enable the tracking of their Ajna token balance by calling `ajnaToken.delegate(address)`. If the passed in `address` argument is that of the caller than this account can now vote by themselves, i.e. delegate to self. If on the other hand a different account's `address` is passed in, that account will receive the voting power of the caller, i.e. delegating to another account. To summarize, until an account calls `ajnaToken.delegate(address)`, that account's tokens cannot be used to place votes on any proposals.

An `ExtraordinaryFunding` proposal can be voted on by everyone that has voting power. Additionally any account can only place a vote once, can only vote in favor of a proposal, can only vote with their entire voting power, and a placed vote cannot be undone. An `ExtraordinaryFunding` proposal succeeds when there are enough votes in favor, and the minimum treshold is adhered to. There is no minimum amount of time that needs to pass before an Extraordinary proposal can be executed after it has gathered enough votes.

## Exploit scenario
Mallory offchain collects a list of all accounts that have voting power. She sums all the voting power and offchain calculates the maximum of Ajna tokens that could be withdrawn if a proposal gathered all of that voting power. Mallory writes a custom contract that, inside the constructor, creates an Extraordinary proposal to transfer all of the Ajna tokens to an account she controls, loops through the collected accounts with voting power and call `GrantFund.voteExtraordinary` for each, followed by a call to `GrantFund.executeExtraordinary`. Mallory deploys the contract and receives the Ajna tokens.

## Proof of Concept

Place the below inside a new file: `ecosystem-coordination/test/DrainGrantFund.sol`
```solidity
pragma solidity 0.8.16;

import { IExtraordinaryFunding } from "../src/grants/interfaces/IExtraordinaryFunding.sol";
[...]

PR #57, opened 2023-03-29
ToB GH issue #55

grandizzy:
- Attacker can use Extraordinary proposal to steal extraordinary amounts of Ajna
- remove `address account_` param of `voteExtraordinary` function, use `msg.sender` instead
- remove `_extraordinaryFundingVote` internal function, logic inline in `voteExtraordinary` function
- add attack test, assert that attacker cannot vote multiple times / on behalf of other addresses
- update unit tests


MikeHathaway (approved):
Looks great! Thanks for the fix

ith-harvey (approved):
`msg.sender` rightfully replaces `account_` LGTM

PR #62, opened 2023-04-04
Revert Ajna token code to deployed version

MikeHathaway:
This reverts the token code to match that of the currently deployed Ajna token on mainnet. 

This can be verified by comparing with the relevant commit: https://github.com/ajna-finance/ecosystem-coordination/blob/15b818a4a9eebf722b2c35760cea998afa095569/src/BaseToken.sol


grandizzy (approved):
looks inline with deployed contract https://etherscan.io/token/0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079#code#F1#L54

prat-gpt (approved):
LGTM!

PR #72, opened 2023-04-30
Forge 1.0 support:

grandizzy:
Invariant tests - in useRandomActor modifier:
- try to start prank using startPrank, if that fails then fallback to changePrank
- remove stopPrank as not needed (changePrank does stop/startPrank)

Unit tests: fix by using startPrank instead changePrank


MikeHathaway (approved):
LGTM! Thanks for the fix

EdNoepel (approved):
Thank you.

PR #78, opened 2023-05-09
Quantstamp: Code Documentation & Best Practices

grandizzy:
* Code documentation:
  - `StandardFunding.sol : readd -> re-add`: changed
  - `ExtraordinaryFunding.sol` shows unclear comments: updated

* Adherence to Best Practices:
  - `StandardFunding._insertionSortProposalsByVotes` should better name the parameter `targetProposalId_` to
better represent its purpose in the function as an index: renamed to `targetProposalIndex_`
  - In the `Maths` library, there is a `abs()` function that returns the absolute size of a signed integer. It will always be positive, so it makes sense to return it as `uint256` instead of `int`: changed, used `SafeCast.toUint256` inside `Maths.abs` instead where function called
  - `Maths.sol` library should use `WAD` instead of hardcoding the powers: updated, made `WAD` constant internal
  - `proposal.startBlock > block.number` check is redundant in `L139` of `ExtraordinaryFunding.sol`: removed
  - Declare `uint256` instead of `uint`: updated all occurrences, same for `int` to `int256`
  - `_standardProposalState() `can be renamed to `_getStandardProposalState()` or viceversa to be consistent
in both contracts: changed function name
  - The error `InsufficientVotingPower` in `L659` of `StandardFunding.sol` is not descriptive for that case. Consider
creating another error with a different name: created `InsufficientRemainingVotingPower` error, updated tests
  - improvements to `StandardFunding.sol`:
    - `L317` The check can be simplified to `sum > _sumProposalFundingVotes(_fundedProposalSlates[currentSlateHash])`: remove redundant check
    - `L448` The check can be moved after the for loop: moved check after loop
    - `L679`, `L849` `SafeCast` is not necessary in these statements: removed with `Maths.abs` change

### DIFF
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -55,7 +55,7 @@ contract GrantFund is IGrantFund, ExtraordinaryFunding, StandardFunding {
     ) external view override returns (ProposalState) {
         FundingMechanism mechanism = findMechanismOfProposal(proposalId_);
 
-        return mechanism == FundingMechanism.Standard ? _standardProposalState(proposalId_) : _getExtraordinaryProposalState(proposalId_);
+        return mechanism == FundingMechanism.Standard ? _getStandardProposalState(proposalId_) : _getExtraordinaryProposalState(proposalId_);
     }
 
     /**************************/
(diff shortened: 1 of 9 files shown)


MikeHathaway (approved):
LGTM

prat-gpt (approved):
LGTM

PR #80, opened 2023-05-10
Q-AJN-6: Inconsistent Time Period

MikeHathaway:
# Description of change
## High level
* Fix inconsistent time period in claimDelegateReward
  * Change operator to `<=` from `<` so as to ensure rewards can't be claimed in the final block of the challenge period.

### DIFF
--- a/src/grants/base/StandardFunding.sol
+++ b/src/grants/base/StandardFunding.sol
@@ -242,7 +242,7 @@ abstract contract StandardFunding is Funding, IStandardFunding {
         QuarterlyDistribution memory currentDistribution = _distributions[distributionId_];
 
         // Check if Challenge Period is still active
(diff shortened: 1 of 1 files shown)


grandizzy (approved):
LGTM (and was flagged by other wardens). Wonder if you could think at a unit test to cover this

prat-gpt (approved):
lgtm!

MikeHathaway:
> LGTM (and was flagged by other wardens). Wonder if you could think at a unit test to cover this

Agree that a test for this would be good, but would be non trivial to implement. Added a card to track as a separate pr: https://www.notion.so/worlds-best-crypto-lending-platform/Create-test-for-Q-AJN-6-2526ec20a02946dfa0bc657fd451d43f?pvs=4

PR #81, opened 2023-05-10
Q-AJN-2: Locked Funds Due to Non-Executable Proposals

MikeHathaway:
# Description of change
## High level
* Validate `recipient` field in `_validateCalldatas` as part of proposal creation, fixing Q-AJN-2.
  * Check that the provided address isn't the 0 address, the Ajna token contract, or the grant fund contract itself. If any of the calldatas have an address set to one of those, the proposal creation will fail in both mechanisms.
  * Add new test `testInvalidRecipientCalldata` to ensure invalid proposal calldatas can't be created.

* Q-AJN-2: Locked Funds Due to Non-Executable Proposals
* Funds could be unreachable if invalid calldata proposals make it through the funding mechanism and can't be executed as the funds won't be returned to the treasury.

### DIFF
--- a/src/grants/base/Funding.sol
+++ b/src/grants/base/Funding.sol
@@ -124,6 +124,8 @@ abstract contract Funding is IFunding, ReentrancyGuard {
             }
             if (selector != bytes4(0xa9059cbb)) revert InvalidProposal();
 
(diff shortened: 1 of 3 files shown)


EdNoepel (approved):
LGTM

grandizzy (approved):
LGTM

PR #82, opened 2023-05-10
Q-AJN-5: Missing Input Validation

MikeHathaway:
# Description of change
## High level
* Add additional external input validations
  * Update `voteExtraordinary`, `fundingVote`, `screeningVote` to check that each successive vote being cast is using non 0 voting power.
  * Update `_validateSlate` to check that a slate contains a proposal, and that each proposal has >0 funding votes received: https://github.com/code-423n4/2023-05-ajna-findings/issues/465
  * Also fixes a bug where a proposal could be executed even if it received no funding votes by changing a check to `<=`
  * Update invariant tests for the new assertions.
  * Add validation to calldata length as suggested in: https://github.com/code-423n4/2023-05-ajna-findings/issues/280

https://github.com/code-423n4/2023-05-ajna-findings/issues/465

### DIFF
--- a/src/grants/base/StandardFunding.sol
+++ b/src/grants/base/StandardFunding.sol
@@ -555,6 +555,9 @@ abstract contract StandardFunding is Funding, IStandardFunding {
             // check that the proposal being voted on is in the top ten screened proposals
             if (_findProposalIndex(voteParams_[i].proposalId, _topTenProposals[currentDistributionId]) == -1) revert InvalidVote();
 
+            // check that the voter isn't attempting to cast a vote with 0 power
+            if (voteParams_[i].votesUsed == 0) revert InvalidVote();
+
             // cast each successive vote
             votesCast_ += _fundingVote(
                 currentDistribution,
@@ -586,6 +589,9 @@ abstract contract StandardFunding is Funding, IStandardFunding {
             if (proposal.distributionId != currentDistribution.id) revert InvalidVote();
 
             uint256 votes = voteParams_[i].votes;
+
+            // check that the voter isn't attempting to cast a vote with 0 power
+            if (votes == 0) revert InvalidVote();
 
             // cast each successive vote
             votesCast_ += votes;
--- a/src/grants/base/StandardFunding.sol
+++ b/src/grants/base/StandardFunding.sol
@@ -374,6 +374,9 @@ abstract contract StandardFunding is Funding, IStandardFunding {
         bytes[] memory calldatas_,
         string memory description_
     ) external override returns (uint256 proposalId_) {
+        // check description string isn't empty
+        if (bytes(description_).length == 0) revert InvalidProposal();
+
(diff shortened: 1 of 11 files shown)


prat-gpt (approved):
LGTM

src/grants/base/StandardFunding.sol
```diff
@@ -427,6 +427,9 @@ abstract contract StandardFunding is Funding, IStandardFunding {
         // check that the slate has no duplicates
         if (_hasDuplicates(proposalIds_)) revert InvalidProposalSlate();
 
+        // check that there is a proposal in the slate
```
grandizzy: suggest better move this at line 428 in same if clause and before checking if it has duplicates

src/grants/base/StandardFunding.sol
```diff
@@ -555,6 +559,9 @@ abstract contract StandardFunding is Funding, IStandardFunding {
             // check that the proposal being voted on is in the top ten screened proposals
             if (_findProposalIndex(voteParams_[i].proposalId, _topTenProposals[currentDistributionId]) == -1) revert InvalidVote();
 
+            // check that the voter isn't attempting to cast a vote with 0 power
+            if (voteParams_[i].votesUsed == 0) revert InvalidVote();
```
grandizzy, line 586: Same as above, better check before findProposalIndex

grandizzy (changes requested):

src/grants/base/StandardFunding.sol
```diff
@@ -438,7 +441,8 @@ abstract contract StandardFunding is Funding, IStandardFunding {
             if (_findProposalIndex(proposalIds_[i], _topTenProposals[distributionId_]) == -1) revert InvalidProposalSlate();
 
             // account for fundingVotesReceived possibly being negative
-            if (proposal.fundingVotesReceived < 0) revert InvalidProposalSlate();
+            // block proposals that recieve no positive funding votes from entering a finalized slate
```
grandizzy, line 457: Missed that one, also better to do the check in same line 441 before even searching the proposal index

ith-harvey (commented):
They suggested that `description_` is checked for non-zero, didn't see that included in PR.

src/grants/base/StandardFunding.sol
MikeHathaway: Updated accordingly

src/grants/base/StandardFunding.sol
MikeHathaway: updated

src/grants/base/StandardFunding.sol
MikeHathaway: updated

MikeHathaway:
> They suggested that `description_` is checked for non-zero, didn't see that included in PR.

Added

grandizzy (approved):

PR #84, opened 2023-05-12
Code Arena 451: add delay to proposeExtraordinary start block

MikeHathaway:
# Description of change
## High level
* Add one block delay to the start block of `proposeExtraordinary`. This makes it more difficult to attackers to manipulate the treasury and token supply temporarily with an eye on securing a successful extraordinary proposal
  * Updates tests.
  * Wasn't able to create a new local variable `startBlock` to reduce duplicate computations due to a stack too deep error.

* https://github.com/code-423n4/2023-05-ajna-findings/issues/451


grandizzy (changes requested):
Suggest including the `testPoCBackRunEFM` test from report so we make sure no longer reproducible

MikeHathaway:
On further review, I believe this isn't a valid attack vector as burning the tokens would simultaneously decrease their voting power and prevent them from being able to cast the requisite votes.

PR #88, opened 2023-05-13
Code Arena: input validation of execute calls

MikeHathaway:
# Description of change
## High level
* Replace user provided values, and targets params with the expected hardcoded values. This saves gas and reduces the attack surface.

* https://github.com/code-423n4/2023-05-ajna-findings/issues/82
* https://github.com/code-423n4/2023-05-ajna-findings/issues/72
* https://github.com/code-423n4/2023-05-ajna-findings/issues/411
* https://github.com/code-423n4/2023-05-ajna-findings/issues/117


grandizzy (commented):
Maybe I am missing something, but why do we still expect the targets and values then in `proposeExtraordinary` / `proposeStandard` if we no longer use them to execute?

MikeHathaway:
> Maybe I am missing something, but why do we still expect the targets and values then in `proposeExtraordinary` / `proposeStandard` if we no longer use them to execute?

Mostly to maintain compliance with other `propose` implementations and streamline external re-usage of the grant fund code. But don't explicitly need it. We would need to add code to fill in hardcoded values for targets and values if we were to replace the usage of external calldata.

prat-gpt (approved):
lgtm. I am in favour of keeping targets and values as mike mentioned for interface compliance with governance contract and also so that we don't have to add hardcoded values for targets and values in the contract. Prefer adding that code on sdk side rather than on the contract side.

grandizzy (approved):
Lgtm, makes sense

PR #64, opened 2023-04-05
TOB-AJNA-7: Prevent collision of proposalIds in standard and extraordinary funding

prat-gpt:
* `ProposalId` collision for extraordinary and standard funding is prevented by hashing strings `Extraordinary Funding: ` and `Standard Funding: ` with description hash respectively

# Gas usage
## Pre Change
```| src/grants/GrantFund.sol:GrantFund contract |                 |        |        |        |         |
|---------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                             | Deployment Size |        |        |        |         |
| 4657644                                     | 23260           |        |        |        |         |
| Function Name                               | min             | avg    | median | max    | # calls |
| claimDelegateReward                         | 1407            | 57405  | 58540  | 65340  | 193     |
| executeExtraordinary                        | 7152            | 42342  | 39567  | 95588  | 8       |
| executeStandard                             | 8775            | 29860  | 37905  | 47659  | 8       |
| findMechanismOfProposal                     | 613             | 2043   | 799    | 4719   | 3       |
| fundTreasury                                | 8001            | 45104  | 44641  | 65872  | 29      |
| fundingVote                                 | 1584            | 105607 | 105266 | 409345 | 208     |
| getDelegateReward                           | 2975            | 3465   | 2975   | 5017   | 5       |
| getDistributionId                           | 374             | 433    | 374    | 2374   | 502     |
| getDistributionPeriodInfo                   | 1046            | 1955   | 1046   | 5046   | 66      |
| getExtraordinaryProposalInfo                | 992             | 992    | 992    | 992    | 3       |
| getExtraordinaryProposalSucceeded           | 2245            | 2664   | 2740   | 3008   | 3       |
| getFundedProposalSlate                      | 1133            | 1309   | 1368   | 1368   | 4       |
| getFundingPowerVotes                        | 15500           | 15688  | 15751  | 15751  | 4       |
| getFundingVotesCast                         | 1577            | 2317   | 2317   | 3057   | 2       |
| getMinimumThresholdPercentage               | 471             | 1227   | 739    | 2471   | 3       |
| getProposalInfo                             | 980             | 980    | 980    | 980    | 105     |
| getSlateHash                                | 872             | 880    | 884    | 884    | 3       |
| getSliceOfNonTreasury                       | 1577            | 4827   | 4827   | 8077   | 2       |
| getSliceOfTreasury                          | 781             | 1781   | 1781   | 2781   | 2       |
[...]

PR #93, opened 2023-05-19
Remove prefix string from proposal description

prat-gpt:
# Description of change
## High level
* Description prefix was added to prevent collision of `proposalIds` of Extraordinary Funding and Standard Funding #64 . Removed prefix string from description since the extraordinary funding mechanism is removed.


grandizzy (approved):

MikeHathaway (approved):
LGTM

PR #94, opened 2023-05-19
Fix Invariants

MikeHathaway:
# Description of change
## High level
* Fix issues with multiple distribution period invariant DP6
* Fix issues with Funding and Finalize scenario timing setup
* Update call validation in handler
* Add additional view functions to enable more elegant access to grant fund temporal state.

### DIFF
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -113,13 +113,13 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
 
     /**
      * @notice Get the block number at which this distribution period's screening stage ends.
-     * @param  startBlock The start block of a distribution period to get the screening stage end block for.
+     * @param  startBlock_ The start block of a distribution period to get the screening stage end block for.
      * @return The block number at which this distribution period's screening stage ends.
     */
     function _getScreeningStageEndBlock(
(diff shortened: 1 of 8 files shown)


grandizzy (approved):
there are still couple of TODOs but happy to have this merged

src/grants/GrantFund.sol
```diff
@@ -1049,6 +1049,11 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
     /*** External View Functions ***/
     /*******************************/
 
+    /// @inheritdoc IGrantFundActions
+    function getChallengeStageStartBlock(uint256 endBlock_) external pure override returns (uint256) {
```
prat-gpt, line 1053: i think we can avoid this method in the contract, although there is no contract size issue. same with `getFundingStageEndBlock` and `getScreeningStageEndBlock`

prat-gpt (approved):
lgtm! added a comment

src/grants/GrantFund.sol
MikeHathaway: We had added those in previously to avoid the usage of "Magic Numbers" and hardcoding the values to reduce likelihood of mistakes. There's definitely a gas cost for doing this.

PR #97, opened 2023-05-25
CodeArena-206: Fix funding stage voting power snapshot block

MikeHathaway:
# Description of change
## High level
* Fix code discrepancy that was using 34 blocks prior to the funding stage start instead of 33 blocks prior to funding stage start for snapshot block as specified in the whitepaper.

* https://github.com/code-423n4/2023-05-ajna-findings/issues/206

### DIFF
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -1010,12 +1010,12 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
         }
         // voter hasn't yet called _castVote in this period
         else {
-            uint256 fundingSTageStartBlock = screeningStageEndBlock_ + 1;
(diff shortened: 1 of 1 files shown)


src/grants/GrantFund.sol
```diff
@@ -1010,12 +1010,15 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
         }
         // voter hasn't yet called _castVote in this period
         else {
+            uint256 fundingSTageStartBlock = screeningStageEndBlock_ + 1;
```
grandizzy: fundingSTageStartBlock typo, looks good otherwise

MikeHathaway: Fixed

grandizzy (approved):

prat-gpt (approved):
LGTM

PR #98, opened 2023-05-25
expand unit test code coverage

MikeHathaway:
# Description of change
## High level
* Expand unit test code coverage.

```
Overall coverage rate after changes:                                                                                                                                                                                             
  lines......: 99.0% (293 of 296 lines)                                                                                                                                                                            
  functions..: 94.9% (56 of 59 functions)                                                                                                                                                                          
  branches...: 98.1% (106 of 108 branches)  
```

* There are some outstanding foundry issues preventing us from reaching 100% code coverage, such as this: https://github.com/foundry-rs/foundry/issues/4310
* The two remaining issues have coverage: `newTopSlate` being `false` is checked on line StandardFunding.t.sol#1151, and `false` case for `_screeningVote` proposal sorting is checked on StandardFunding.t.sol#819

### DIFF
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -572,14 +572,14 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
         for (uint256 i = 0; i < numProposalsInSlate_; ) {
             Proposal memory proposal = _proposals[proposalIds_[i]];
 
-            // account for fundingVotesReceived possibly being negative
-            // block proposals that recieve no positive funding votes from entering a finalized slate
-            if (proposal.fundingVotesReceived <= 0) revert InvalidProposalSlate();
-
             // check if Proposal is in the topTenProposals list
             if (
                 _findProposalIndex(proposalIds_[i], _topTenProposals[distributionId_]) == -1
             ) revert InvalidProposalSlate();
+
+            // account for fundingVotesReceived possibly being negative
+            // block proposals that recieve no positive funding votes from entering a finalized slate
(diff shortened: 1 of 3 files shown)


EdNoepel (approved):
LGTM

src/grants/GrantFund.sol
```diff
@@ -572,15 +572,15 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
...
-            if (proposal.fundingVotesReceived <= 0) revert InvalidProposalSlate();
-
             // check if Proposal is in the topTenProposals list
             if (
                 _findProposalIndex(proposalIds_[i], _topTenProposals[distributionId_]) == -1
             ) revert InvalidProposalSlate();
 
+            // account for fundingVotesReceived possibly being negative
+            // block proposals that recieve no positive funding votes from entering a finalized slate
+            if (proposal.fundingVotesReceived <= 0) revert InvalidProposalSlate();
```
grandizzy, line 582: is this change done only for improving coverage or is the same if we keep same sequence of checking? For ux / gas POV would make more sense to keep it as it is and not call _findProposalIndex if no or negative votes received...

grandizzy (commented):
Looks good but wonder how much switching the sequence of proposal validation matters for test coverage

src/grants/GrantFund.sol
MikeHathaway: Making this change improved coverage, as the other case could never be reached in a unit test as it would always trigger the unit test

grandizzy (approved):

PR #101, opened 2023-06-01
CodeArena-288: Incorrect Voting Power

MikeHathaway:
# Description of change
## High level
* Fix bug whereby someone would be able to have voting power even if they didn't keep the tokens for the full snapshot period if they simply transferred the tokens to the account on the distribution start block.
* Add additional unit test proving the fix.

* https://github.com/code-423n4/2023-05-ajna-findings/issues/288


grandizzy (approved):
LGTM

EdNoepel (approved):
LGTM

PR #110, opened 2023-06-19
Invariant Improvements: Fix Actor log and improve `screeningVote` handler

prat-gpt:
# Description of change
## High level
* `logActorSummary` was reverting with `underflow/overflow` when no distribution is started. Fixed with starting distribution in `setUp`
* Improved `screeningVote` handler to vote for multiple proposals by generating parameters with `_screeningVoteParams`

### DIFF
--- a/test/invariants/scenarios/MultipleDistributionInvariant.t.sol
+++ b/test/invariants/scenarios/MultipleDistributionInvariant.t.sol
@@ -17,6 +17,8 @@ contract MultipleDistributionInvariant is StandardTestBase {
     function setUp() public override {
         super.setUp();
 
+        startDistributionPeriod();
+
         // set the list of function selectors to run
         bytes4[] memory selectors = new bytes4[](8);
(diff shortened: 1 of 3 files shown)


test/invariants/scenarios/MultipleDistributionInvariant.t.sol
```diff
@@ -17,6 +17,8 @@ contract MultipleDistributionInvariant is StandardTestBase {
     function setUp() public override {
         super.setUp();
 
+        startDistributionPeriod();
```
MikeHathaway: Think it would be better to just wrap the `logActorSummary` in `if (distributionId != 0)`, since wanted to test the process of starting a distribution period in invariants not have it done automatically.

MikeHathaway (approved):
LGTM - good to merge after resolving conflicts. Can remove the logging change in this PR. Thank you for finding that issue

PR #113, opened 2023-06-21
Invariant Improvements: Invariant CS7

prat-gpt:
# Description of change
## High level
* Add invariant CS7: The highest submitted funded proposal slate should have won or tied depending on when it was submitted.


MikeHathaway (approved):
LGTM

PR #116, opened 2023-06-22
Fuzz Test improvements: Fuzz test for screening and funding stage

prat-gpt:
# Description of change
## High level
* Add fuzz tests specific for the screening and funding stage where a random voter votes a random amount of votes on a random proposal random no of times. 
* This test checks the voter's voting power is updated correctly, the proposals vote received is updated correctly and the top ten proposals are filtered correctly.


MikeHathaway (approved):
LGTM! Thanks for the extra tests

EdNoepel (approved):
Interesting that the number of blocks to roll were not fuzzed or otherwise randomized.  Still, a good new test.

PR #120, opened 2023-06-26
Sherlock-79: Burn ajna tokens on wrap

MikeHathaway:
# Description of change
## High level
* Instead of transferring Ajna to the BurnWrapper contract from which they can't be withdrawn, instead burn ajna tokens on wrap.

* https://github.com/sherlock-audit/2023-04-ajna-judging/issues/79
* This isn't really a bug or a vulnerability so much as an alternative design approach. It has some cons as well, as the burn wrapped tokens are potentially usable on other chains still making the total supply potentially less clear.

### DIFF
--- a/src/token/BurnWrapper.sol
+++ b/src/token/BurnWrapper.sol
@@ -12,11 +12,11 @@ import { ERC20Wrapper }   from "@oz/token/ERC20/extensions/ERC20Wrapper.sol";
 import { IERC20Metadata } from "@oz/token/ERC20/extensions/IERC20Metadata.sol";
 
 /**
-    * @title Ajna Token `ERC20` token interface.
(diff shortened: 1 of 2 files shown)


EdNoepel (approved):
The question seems to be:
**When tokens are moved from mainnet, do we want total supply to drop?**  I lean towards "yes", primarily because we can no longer track the token off mainnet.  We do not want total supply to track tokens which may be bridged and later burned on the L2.

ith-harvey (approved):

prat-gpt (approved):
LGTM.

MikeHathaway:
Updated natspec if you want to take another look @EdNoepel

EdNoepel (approved):
LGTM

grandizzy (approved):

PR #123, opened 2023-08-01
Deploy BurnWrappedAjna

EdNoepel:
# Description of change
## High level
Implement forge script for deploying `BurnWrappedAjna` contract.  Could not do this with `forge create` because build process did not like different compiler versions.  Not happy about needing to manually edit source of `BurnWrapper.sol` such that it validates against the correct token address.

No changes to `src/`.

See alternate implementation PR#124 for an implementation which edits `src/` to remove the token address validation.


MikeHathaway (approved):
LGTM

PR #129, opened 2023-08-15
Invariant Improvements: Add invariant scenario for multiple fund treasury between distributions

prat-gpt:
# Description of change
## High level
* Add invariant scenario with multiple treasury funding, transfer ajna between actors and new actors being added.

### DIFF
--- a/test/invariants/handlers/Handler.sol
+++ b/test/invariants/handlers/Handler.sol
@@ -70,7 +70,10 @@ contract Handler is Test, GrantFundTestHelper {
         _tokenDeployer = tokenDeployer_;
 
         // instantiate actors
-        actors = _buildActors(numOfActors_, tokensToDistribute_);
+        address[] memory newActors = _buildActors(numOfActors_, tokensToDistribute_);
+        for (uint256 i = 0; i < newActors.length; ++i) {
+            if (newActors[i] != address(0)) actors.push(newActors[i]);
+        }
 
         // set Test invariant contract
         testContract = ITestBase(testContract_);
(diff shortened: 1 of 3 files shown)


test/invariants/handlers/Handler.sol
```diff
@@ -70,7 +70,10 @@ contract Handler is Test, GrantFundTestHelper {
         _tokenDeployer = tokenDeployer_;
 
         // instantiate actors
-        actors = _buildActors(numOfActors_, tokensToDistribute_);
+        address[] memory newActors = _buildActors(numOfActors_, tokensToDistribute_);
+        for (uint256 i = 0; i < newActors.length; ++i) {
+            if (newActors[i] != address(0)) actors.push(newActors[i]);
```
MikeHathaway, line 75: Were any of the newActors addresses actually `0x0`?

MikeHathaway (approved):
LGTM

test/invariants/handlers/Handler.sol
prat-gpt: yes, i got some failures where actor addresses were `0x0`, it was happening [here](https://github.com/ajna-finance/ecosystem-coordination/blob/16f0cfe6e49622470333277815d91ae91c327e61/test/invariants/handlers/Handler.sol#L155).

EdNoepel (approved):
LGTM

test/invariants/handlers/Handler.sol
MikeHathaway: Ah, yeah from the failure path. Makes sense

PR #132, opened 2023-09-15
return 0 for getDelegateRewards if voter didn't vote in the screening stage

MikeHathaway:
# Description of change
## High level
* return 0 for `getDelegateRewards` if voter didn't vote in the screening stage

* `getDelegateReward` returned incorrect values if called when a voter didn't vote in the screening stage.
* https://github.com/Fixed-Point-Solutions/ajna-PositionManager-RewardsManager/issues/24

### DIFF
--- a/src/grants/GrantFund.sol
+++ b/src/grants/GrantFund.sol
@@ -1099,6 +1099,8 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
         DistributionPeriod storage currentDistribution = _distributions[distributionId_];
         VoterInfo          storage voter               = _voterInfo[distributionId_][voter_];
 
+        if (voter.screeningVotesCast == 0) return 0;
+
         rewards_ = _getDelegateReward(currentDistribution, voter);
     }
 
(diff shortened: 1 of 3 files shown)


grandizzy (commented):
shouldn't this check be done inside `_getDelegateReward` function so to be applied for all scenarios where it is called?

src/grants/GrantFund.sol
```diff
@@ -1099,6 +1099,8 @@ contract GrantFund is IGrantFund, Storage, ReentrancyGuard {
         DistributionPeriod storage currentDistribution = _distributions[distributionId_];
         VoterInfo          storage voter               = _voterInfo[distributionId_][voter_];
 
+        if (voter.screeningVotesCast == 0) return 0;
```
grandizzy, line 1102: I personally prefer using
```Solidity
if (voter.screeningVotesCast != 0) {
  rewards_ = 
}
```

MikeHathaway:
> shouldn't this check be done inside `_getDelegateReward` function so to be applied for all scenarios where it is called?

Did it this way for a couple of reasons - this would be the minimal amount of changes, and also, the claim approach should revert instead of returning 0, so we would have to create branching logic to handle claim vs get calls to `_getDelegateReward`

grandizzy:
> > shouldn't this check be done inside `_getDelegateReward` function so to be applied for all scenarios where it is called?
> 
> Did it this way for a couple of reasons - this would be the minimal amount of changes, and also, the claim approach should revert instead of returning 0, so we would have to create branching logic to handle claim vs get calls to `_getDelegateReward`

makes sense, thanks

EdNoepel (approved):
LGTM

src/grants/GrantFund.sol
ith-harvey: Although I'm not opinionated about this change, it's a pattern we have been following in the contracts repo and therefore I agree with @grandizzy .

ith-harvey (approved):
LGTM

prat-gpt (approved):
LGTM

